PR #1, opened 2022-02-03
fix(script-dev): add restart flag when dev script fails

djcouto:
## Description

The script `npm run dev` was failing sometimes because the storybook preview requires the file `dist/esm/loader.js` which is only available after stencil finishes compiling the components.

This PR fixes this by adding a restart flag and a restart delay timeout.


miguelpinto98 (approved):

package.json
```diff
@@ -1,54 +1,54 @@
...
-  "scripts": {
-    "build": "stencil build --docs",
-    "start": "stencil build --dev --watch --serve",
-    "test": "stencil test --spec --e2e",
-    "test.watch": "stencil test --spec --e2e --watchAll",
-    "generate": "stencil generate",
-    "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook",
-    "dev:stencil": "stencil build --watch",
-    "dev": "run-p dev:stencil storybook",
```
ricardogoncalves89, line 30: Probably you can remove the `npm-run-all` dev-dependency. please check it

ricardogoncalves89 (commented):
It would be nice if you update the README accordingly to the new scripts

package.json
djcouto: Yes, we most certainly can, will remove it on the upcoming commit.

PR #2, opened 2022-02-09
feat: bootstrap the theme (spacing, colors and typography) [SA-593]

ricardogoncalves89:
- Change PeacockUI to use Sass
- Architect the CSS Variables for the color theme 
- Define typography
- Define spacial system


pedrorebelocosta (approved):

src/styles/pk-spacing.scss
```diff
@@ -0,0 +1,13 @@
...
+
+$spacing-xxs: 2px;
+$spacing-xs: 4px;
+$spacing-s: 8px;
+$spacing-m: 12px;
+$spacing-l: 16px;
+$spacing-xl: 20px;
+$spacing-xxl: 24px;
+$spacing-xxxl: 32px;
+$spacing-xxxxl: 64px;
```
paulosoaresvieira, line 13: we now have more spacings... since they are on the "middle" i'd suggest to add them here

paulosoaresvieira (commented):
change the pk names and add the new spacings

src/styles/pk-spacing.scss
ricardogoncalves89: I will update it. We probably need to change the naming. Any suggestion?

djcouto (approved):

paulosoaresvieira (approved):

PR #5, opened 2022-02-11
feat(switch-button): add component

djcouto:
#### Demo
Screenshot 2022-02-16 at 10 42 30

### DIFF
--- a/packages/react-ui-components/src/stories/Components/ActionButton.stories.tsx
+++ b/packages/react-ui-components/src/stories/Components/ActionButton.stories.tsx
@@ -3,7 +3,7 @@ import React from 'react';
 import { EActionButtonType, KvActionButton } from '../../components';
 
 export default {
-	title: "Components/Buttons",
+	title: "Components/Buttons/Action",
 	component: KvActionButton
 } as ComponentMeta<typeof KvActionButton>;
 
--- a/packages/ui-components/src/types.ts
+++ b/packages/ui-components/src/types.ts
@@ -1,2 +1,3 @@
 export * from './components';
 export * from './components/action-button/action-button.types';
+export * from './components/switch-button/switch-button.types';
--- a/packages/ui-components/stencil.config.ts
+++ b/packages/ui-components/stencil.config.ts
@@ -3,14 +3,14 @@ import { Config } from '@stencil/core';
 import { reactOutputTarget as react } from '@stencil/react-output-target';
 import { sass } from '@stencil/sass';
 
+// Configuration to generate properties with 2 way bind in angular
 const angularValueAccessorBindings: ValueAccessorConfig[] = [
-	// Configuration example to generate properties with 2 way bind in angular
-	// {
-	// 	elementSelectors: ['component-name'],
-	// 	event: 'eventNameOfChamge',
-	// 	targetAttr: 'propName',
-	// 	type: 'number/type'
-	// }
+	{
+		elementSelectors: ['kv-switch-button'],
+		event: 'switchStateChange',
+		targetAttr: 'state',
+		type: 'text'
+	}
 ];
 
 export const config: Config = {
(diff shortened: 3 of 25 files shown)


pedrorebelocosta (approved):

packages/react-ui-components/package.json
```diff
@@ -32,6 +32,7 @@
 		"@storybook/builder-webpack5": "^6.4.19",
 		"@storybook/manager-webpack5": "^6.4.19",
 		"@storybook/react": "^6.4.19",
+		"@types/estree": "^0.0.50",
```
miguelpinto98, line 35: rollback this please. Is unnecessary since the rollback that happen now.

packages/ui-components/src/types.ts
```diff
@@ -1,2 +1,3 @@
 export * from './components';
 export * from './components/action-button/action-button.types';
+export * from './components/switch-button/switch-button.types';
```
miguelpinto98, line 3: I think we need also export this interface in react application `src/components/index`. 
Also declare the component inside componeent-library.module.ts (angular lib) to export the component.

If you want to enable two way binding in `state` prop for angular, you can use `angularValueAccessorBindings` configuration inside stencil.config.js

miguelpinto98 (changes requested):
Check my comments, it's only to expose the component for react/angular libs. 
Also create a story inside the react library.

packages/ui-components/stencil.config.ts
```diff
@@ -3,14 +3,14 @@ import { Config } from '@stencil/core';
...
-	// 	elementSelectors: ['component-name'],
-	// 	event: 'eventNameOfChamge',
-	// 	targetAttr: 'propName',
-	// 	type: 'number/type'
-	// }
+	{
+		elementSelectors: ['kv-switch-button'],
+		event: 'switchStateChange',
+		targetAttr: 'state',
+		type: 'text'
```
djcouto, line 12: @miguelpinto98 

Since this is an enum it could make sense to change but sincerely I don't know the implications of changing it. Do you know?

miguelpinto98: I think is related with HTML elements types, like radio, select, checkbox, number, text... And convert it to it. We can go with this, and text a different approach if we need.

packages/react-ui-components/src/stories/Components/ActionButton.stories.tsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import { EActionButtonType, KvActionButton } from '../../components';
 
 export default {
-	title: "Components/Buttons",
+	title: "Components/Buttons/Action",
```
miguelpinto98, line 6: Action should be added?

djcouto: This is more a design choice 😄

The purpose of this change was to have all buttons aggregated under the same `Button` folder as you can see on this [screenshot](https://user-images.githubusercontent.com/17494742/154248291-e6a1c007-e4fb-4db7-b401-151ef39ae879.png).

```
- Buttons
   - ActionButton
   - SwitchButton
```

But I can easily rollback here if you don't think it makes sense 😄

packages/ui-components/stencil.config.ts
djcouto: Ok gotcha, thanks 👍

packages/react-ui-components/src/stories/Components/ActionButton.stories.tsx
miguelpinto98: I didn't understand until now. 😄  Let's continue with this.

miguelpinto98 (approved):
🚀

ricardogoncalves89 (approved):
👌

ricardogoncalves89 (approved):
great work with the Tests 👌

PR #36, opened 2022-03-04
feat(colors-palette): add missing neutrals to colors palette

djcouto:


packages/ui-components/src/assets/styles/kv-theme.default.scss
```diff
@@ -22,22 +22,24 @@ $warning-color: #ffc043 !default;
...
 // `-rgb` prop is used to apply transparencies with RGBA
-$neutral-color-1: mix($text-color, $background-color, 6%);
-$neutral-color-2: mix($text-color, $background-color, 11.5%);
-$neutral-color-3: mix($text-color, $background-color, 15.1%);
-$neutral-color-4: mix($text-color, $background-color, 29%) !default;
-$neutral-color-5: mix($text-color, $background-color, 64%) !default;
-$neutral-color-6: mix($text-color, $background-color, 86%) !default;
+$neutral-color-0: #fff;
+$neutral-color-9: #121212;
+$neutral-color-1: mix($neutral-color-9, $neutral-color-0, 5.5%) !default; // #f2f2f2
```
ricardogoncalves89, line 35: You didn't need calculate the colors with `mix` the color can be fixed this is only a helper if we swap the base colors

packages/ui-components/src/assets/styles/kv-theme.default.scss
```diff
@@ -47,56 +49,53 @@ $colors: (
...
 	'error': (
-		base: $error-color, 
-		contrast: #fff, 
-		dark: get-color-dark($error-color), 
+		base: $error-color,
+		contrast: #fff,
+		dark: get-color-dark($error-color),
 		light: get-color-light($error-color)
 	),
-	'background': (
```
ricardogoncalves89, line 95: Probably this is breaking. Why do you remove?

ricardogoncalves89 (changes requested):

packages/ui-components/src/assets/styles/kv-theme.default.scss
djcouto: Yes, I see your point. I meant to leave these calculations here just in case we might need them in the future. I'll manually set all colours.

ricardogoncalves89 (approved):

pedrorebelocosta (approved):

PR #35, opened 2022-03-04
feat: add icons and illustrations

ricardogoncalves89:
*Missing unit and e2e tests
image
image
image

PR #37, opened 2022-03-04
fix(switch-button): add night mode style

djcouto:
##### Default
Screenshot 2022-03-04 at 09 54 26

##### Disabled
Screenshot 2022-03-04 at 18 06 28


packages/ui-components/src/components/switch-button/switch-button.scss
```diff
@@ -1,11 +1,39 @@
...
-$sqr-radius: 2px;
-$icon-size: 16px;
+$btn-height: 20px;
+$btn-width: 40px;
+$sqr-size: 16px;
+$sqr-padding-x: 2px;
+$icon-size: 12px;
+
+@function get-sqr-x-translation() {
+	@return $btn-width - $sqr-size - $sqr-padding-x * 2;
```
ricardogoncalves89, line 10: If I set the `--btn-height` var the calculation don't return the correct value. 

Suggestion: 
```
:host {
	--btn-height: 20px;

	$btn-height: var(--btn-height);
}
```
after that use only the `$btn-height` for styling and calculations

pedrorebelocosta:
This is pending as it requires https://github.com/kelvininc/ui-components/pull/35 and https://github.com/kelvininc/ui-components/pull/36 to be merged right?

ricardogoncalves89 (changes requested):

packages/ui-components/src/components/switch-button/switch-button.scss
djcouto: You're absolutely correct.

Fixing on the upcoming commit.

ricardogoncalves89 (approved):
great work with size styling 👌

pedrorebelocosta (approved):

pedrorebelocosta:
top notch work :D

PR #45, opened 2022-03-08
feat(svg-icon): enable public assets url configuration

djcouto:
#### Description
Currently, the links to the SVG icons are relative to the page, which implies the SVG icons file to be served on every relative page where it's running.
This PR enables this public assets URL to be configured via prop and also defaults it to be absolute.

### DIFF
--- a/packages/ui-components/src/components/svg-icon/readme.md
+++ b/packages/ui-components/src/components/svg-icon/readme.md
@@ -88,6 +88,7 @@ export const SvgIconExample: React.FC = () => (
 
 | Property            | Attribute      | Description                                                                                                                 | Type                 | Default     |
 | ------------------- | -------------- | --------------------------------------------------------------------------------------------------------------------------- | -------------------- | ----------- |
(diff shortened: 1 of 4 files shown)


packages/ui-components/src/components/svg-icon/readme.md
```diff
@@ -100,22 +101,6 @@ export const SvgIconExample: React.FC = () => (
 | `"icon"` | The icon container. |
 
 
-## CSS Custom Properties
```
ricardogoncalves89, line 103: why is this removed?

ricardogoncalves89 (changes requested):

packages/ui-components/src/components/svg-icon/readme.md
djcouto: Ups, don't know, I didn't notice, probably something went wrong when generating the readme.

I will reset this on the upcoming commit.

ricardogoncalves89 (approved):

pedrorebelocosta (approved):

PR #49, opened 2022-03-11
feat(summary-card): add summary card component

pedrorebelocosta:
>image

>image

### DIFF
--- a/packages/angular-ui-components/src/kv-ui-components.module.ts
+++ b/packages/angular-ui-components/src/kv-ui-components.module.ts
@@ -20,7 +20,8 @@ import {
 	KvTabNavigation,
 	KvTagLetter,
 	KvTextField,
-	KvTooltip
+	KvTooltip,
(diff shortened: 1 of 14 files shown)


miguelpinto98:
I don't know if we gonna continue using this card, since the cards has problem with larger descriptions. Or probably turn this card more rectangular.

pedrorebelocosta:
> I don't know if we gonna continue using this card, since the cards has problem with larger descriptions. Or probably turn this card more rectangular.

Yeah... I've been messing around trying to make it styled with an ellipsis which you can hover and get a tooltip for the remainder of the text, obviously not the best solution but it could work.

miguelpinto98 (approved):
Seems good! But I still have some concerns related with the card design. But if Design wants with for the future, it's ok for me.

djcouto (approved):

ricardoarasilva (approved):

PR #59, opened 2022-03-17
feat(illustration): add component

djcouto:
This PR separates the illustrations svgs to their own component `kv-illustration`.

Since these svgs deal with a lot o mixed colors with no apparent direct correlation between light and dark mode, I decided to move all illustrations into separated `*-night` and `*-light` symbols.

Apart from the `pk-es-section-somethingwentwrong` and `kv-es-table-search`, all night illustrations are the same as the light ones.


packages/ui-components/src/assets/svg-symbols.svg
```diff
@@ -1147,14 +1147,14 @@
 	<symbol id="kv-es-section-somethingwentwrong"
```
miguelpinto98, line 1147: this illustration works with light theme? Probably some colors needs to be defined in a variable.

miguelpinto98 (changes requested):

ricardoarasilva (approved):

miguelpinto98 (approved):

PR #63, opened 2022-03-22
feat: Search component

GoncaloCamaz:


ricardoarasilva (changes requested):
Hi @GoncaloCamaz needed to add the new component to the angular library.

GoncaloCamaz:
Screenshot 2022-03-22 at 18 59 24
Screenshot 2022-03-22 at 19 00 01
Screenshot 2022-03-22 at 19 00 23

packages/ui-components/src/components/search/usage/angular.md
```diff
@@ -0,0 +1,10 @@
+```html
+<!-- Default -->
+<kv-search></kv-search>
+
+<!-- Labeled -->
+<kv-search label="Text Field"></kv-search>
```
djcouto, line 6: don't think that this prop exists.

also it's missing the `placeholder` and `size` example

packages/ui-components/src/globals/globals.ts
```diff
@@ -23,7 +23,9 @@ export const initialize = (userConfig: UIComponentsConfig = {}) => {
 					console.warn(`Invalid kv-component mode: ${elmMode}, expected: ${Object.values(StyleMode).join(',')}`);
 				}
 			}
-			elm = elm.parentElement;
+			// TODO: check with more time why cant obtain the parentElement
+			// elm = elm.parentElement;
+			elm = doc.body;
```
djcouto, line 28: Was this intended?

djcouto (changes requested):
Good work 💪
Left a few comments 👍

packages/ui-components/src/components/text-field/text-field.tsx
```diff
@@ -126,6 +131,9 @@ export class KvTextField {
 										}}
 									/>
 								)}
+								<div class={{ 'right-slot-container': true, 'focus': this.focused }}>
```
miguelpinto98, line 134: Only to understand the tech decision... Why we decide to use a slot here to draw anything (like KvIcon)?

packages/ui-components/src/components/search/search.tsx
```diff
@@ -0,0 +1,66 @@
...
+	/** (optional) Sets this tab item to a different styling configuration */
+	@Prop() size?: EComponentSize = EComponentSize.Large;
+	/** (optional) Search disabled */
+	@Prop({ reflect: true }) disabled = false;
+	/** Search value*/
+	@Prop({ reflect: true }) value: string = '';
+	/** Emitted when search text value changes */
+	@Event() textChange: EventEmitter<string>;
+	/** Emitted when text field lost focus */
+	@Event() textFieldBlur: EventEmitter<string>;
```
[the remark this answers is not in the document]
GoncaloCamaz: Will add interfaces to the events on next commit

packages/ui-components/src/globals/globals.ts
GoncaloCamaz: This was a fix requested by @ricardogoncalves89 to address a bug. It's a temporary fix.

packages/ui-components/src/components/text-field/text-field.tsx
GoncaloCamaz: @ricardogoncalves89

packages/ui-components/src/components/search/search.tsx
```diff
@@ -0,0 +1,66 @@
...
+})
+export class KvSearch {
+	/** (optional) Search place holder */
+	@Prop({ reflect: true }) placeholder: string = 'Search';
+	/** (optional) Sets this tab item to a different styling configuration */
+	@Prop() size?: EComponentSize = EComponentSize.Large;
+	/** (optional) Search disabled */
+	@Prop({ reflect: true }) disabled = false;
+	/** Search value*/
+	@Prop({ reflect: true }) value: string = '';
```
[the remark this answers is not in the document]
GoncaloCamaz: Will be fixed on next commit

packages/ui-components/src/components/text-field/text-field.tsx
GoncaloCamaz: @miguelpinto98 , the reason to use the slot was taken in order to be able to re-use the textfield with other components like the dropdown. With this solution, the logic behind opening the dropdown or clicking the reset button can stay on the Search/Dropdown component

packages/ui-components/src/components/search/search.types.ts
```diff
@@ -0,0 +1,10 @@
+import { EventEmitter } from '@stencil/core';
```
miguelpinto98, line 1: this interface is used in somewhere else?

packages/ui-components/src/components/text-field/text-field.base.scss
```diff
@@ -170,6 +176,34 @@ input {
...
+	.right-slot-container {
+		position: absolute;
+		top: 0;
+		right: 0;
+		height: 100%;
+		width: $right-slot-width;
+		overflow: hidden;
+		
+		::slotted(kv-svg-icon) {
+			--icon-width: 16px;
```
miguelpinto98, line 189: we need to override the default configuration with the same values?

miguelpinto98 (commented):
Practically good to go, but a 2 questions before it. ^

packages/react-ui-components/src/stories/Components/Search.stories.tsx
```diff
@@ -0,0 +1,47 @@
+import { ComponentStory } from '@storybook/react';
+import React from 'react';
+import { EComponentSize, KvSearch, KvTextField } from '../../components';
+
+KvTextField.displayName = 'KvSearch';
+
+export default {
+	title: 'Inputs/Search',
+	component: 'kv-search',
+	argTypes: {
```
miguelpinto98, line 10: It's possible to add here the events that are generated by the component?
I think it's cool to have this. 

[Look Here](https://storybook.js.org/docs/react/essentials/actions)

packages/ui-components/src/components/text-field/text-field.base.scss
GoncaloCamaz: The css added is necessary in order to place the icon. After checking the possibility of removing the repeated properties, the icon would move out of place.

packages/react-ui-components/src/stories/Components/Search.stories.tsx
GoncaloCamaz: It is possible! But every component is lacking this feature. 
@ricardogoncalves89

packages/ui-components/src/components/search/search.tsx
```diff
@@ -0,0 +1,69 @@
...
+	private searchIcon = 'kv-search';
+	private resetIcon = 'kv-close';
+
+	private onTextChange = (event: CustomEvent<string>) => {
+		this._value = event.detail;
+		this.showResetIcon = !isEmpty(this._value);
+	};
+
+	private onResetClick = (event: MouseEvent) => {
+		this.showResetIcon = false;
```
miguelpinto98, line 47: This value gonna change when the user clean the result (if they want it) if not we gonna have a state problem since the X will disappear but the text value continue with the same data.

GoncaloCamaz: The button now appears if the value has any text on it. The value itself is updated from the parent component of the search via prop. 
The state showResetIcon was removed.

miguelpinto98 (approved):
🚀

packages/ui-components/src/components/text-field/text-field.tsx
djcouto: Not a huge fan of the approach. Since the `right-slot` will always be an icon, we could pass down only the icon name and emit an event when the icon is clicked, which would be then handled by the parent. The end result is the same and less error prone.

With that being said, I see the point of having a fully customizable right slot in which his behaviour is not handled by the text-field component as it should not. 

For that sake, I'll not oppose this approach, just wanted to share with you an alternative way 👍

packages/ui-components/src/components/text-field/text-field.base.scss
```diff
@@ -1,5 +1,7 @@
 @import '../../assets/styles/globals';
 
+$right-slot-width: 36px;
```
djcouto, line 3: This variable should be customizable, please move it to the `:host` selector as a CSS property with this value as default.

packages/ui-components/src/components/text-field/text-field.tsx
```diff
@@ -87,6 +87,10 @@ export class KvTextField {
 		return Array.isArray(value) ? value : [value];
 	}
 
+	private get hasRightSlot() {
+		return !!this.el.querySelector('[slot="right-slot"]');
+	}
```
djcouto, line 92: 👍

djcouto (changes requested):
Great work, these are my final comments and suggestions 💪

If you need any further explanation please ping me 👍

packages/ui-components/src/components/search/search.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { isEmpty } from 'lodash-es';
+import { EComponentSize } from '../../types';
+import { EInputFieldType } from '../text-field/text-field.types';
+import { ISearchEvents } from './search.types';
+
+@Component({
+	tag: 'kv-search',
+	shadow: true
+})
+export class KvSearch implements ISearchEvents {
```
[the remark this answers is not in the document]
GoncaloCamaz: Thank you for the suggestion, I will change it to implements ITextFieldEvents, ISearchEvents and remove the two events from ISearchEvents that are also on ITextFieldEvents.

ricardoarasilva (approved):

packages/ui-components/src/components/search/test/search.e2e.ts
```diff
@@ -0,0 +1,31 @@
...
+describe('Search (end-to-end)', () => {
+	let page: E2EPage;
+
+	describe('when renders with default props', () => {
+		beforeEach(async () => {
+			page = await newE2EPage();
+			await page.setContent('<kv-search></kv-search>');
+		});
+
+		describe('when blur event is emitted', () => {
```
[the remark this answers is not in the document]
GoncaloCamaz: Added e2e tests

packages/ui-components/src/components/search/test/search.spec.tsx
```diff
@@ -0,0 +1,48 @@
...
+		it('should initialize `disabled` with false', () => {
+			expect(component.disabled).toBe(false);
+		});
+
+		it('should initialize `size` with large', () => {
+			expect(component.size).toBe(EComponentSize.Large);
+		});
+	});
+
+	describe('when is disabled', () => {
```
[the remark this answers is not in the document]
GoncaloCamaz: Added missing test

packages/ui-components/src/components/text-field/text-field.base.scss
GoncaloCamaz: Setted as default variable

djcouto (approved):
🚀

PR #65, opened 2022-03-24
feat(kv-radio-button-group): create new components kv-radio-button and kv-radio-button-group

ricardoarasilva:
…adio and create new components kv-radio-button and kv-radio-button-group

### DIFF
--- a/packages/ui-components/src/components/radio-button-group/radio-button-group.tsx
+++ b/packages/ui-components/src/components/radio-button-group/radio-button-group.tsx
@@ -1,0 +1,32 @@
+import { Component, h, Prop, Host, Listen, Event, EventEmitter } from '@stencil/core';
+import { IRadioButton } from '../radio-button/radio-button.types';
+
+@Component({
+	tag: 'kv-radio-button-group',
+	styleUrl: 'radio-button-group.scss',
+	shadow: true
+})
+export class KvRadioButtonGroup {
+	/** (required) List of radio buttons */
+	@Prop({ reflect: true }) buttons: { [key: string]: IRadioButton } = {};
+
+	/** When the radio button selection changes, emit the requested tab's key */
+	@Event() radioButtonChange: EventEmitter<string>;
+
+	/** Listen to custom DOM event of radio button selection */
+	@Listen('checkedChange')
+	radioChangeHandler(event: CustomEvent<string>) {
+		this.buttons[event.detail].active = !this.buttons[event.detail].active;
+		this.radioButtonChange.emit(event.detail);
+	}
+
+	render() {
+		return (
+			<Host>
+				{Object.keys(this.buttons).map(key => (
+					<kv-radio-button label={this.buttons[key].label} id={key} disabled={this.buttons[key].disabled} checked={this.buttons[key].active}></kv-radio-button>
+				))}
+			</Host>
+		);
+	}
+}
--- a/packages/ui-components/src/components/radio-button/radio-button.scss
+++ b/packages/ui-components/src/components/radio-button/radio-button.scss
@@ -1,0 +1,68 @@
+@import '../../assets/styles/globals';
+
+:host {
+	/**
+	 * @prop --button-height: Radio button component's height.
+	 * @prop --text-color-default: Radio button component's text color default.
+	 * @prop --text-color-active: Radio button component's text color active.
+	 * @prop --text-color-disabled: Radio button component's text color disabled.
+	 * @prop --background-color-default: Radio button component's background color default.
+	 * @prop --background-color-active: Radio button component's background color active.
+	 * @prop --background-color-disabled: Radio button component's background color disabled.
+	 * @prop --border-color-default: Radio button component's border color default.
+	 * @prop --border-color-active: Radio button component's border color active.
+	 * @prop --border-color-disabled: Radio button component's border color disabled.
+	 */
+
+	--button-height: 36px;
+	--text-color-default: #{kv-color('neutral-2')};
+	--text-color-active: #{kv-color('neutral-2')};
+	--text-color-disabled: #{kv-color('neutral-4')};
+	--background-color-default: #{linear-gradient(45deg, kv-color('neutral-6') 50%, kv-color('neutral-6') 50%)};
+	--background-color-active: #{linear-gradient(45deg, kv-color('neutral-6') 50%, kv-color('neutral-6') 50%)};
+	--background-color-disabled: #{linear-gradient(45deg, kv-color('neutral-6') 50%, kv-color('neutral-6') 50%)};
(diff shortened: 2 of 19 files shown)


miguelpinto98:
This PR should be divided into 2. 
- The first one to the refactor of `kv-radio-button` to `kv-radio`;
- The new `kv-radio-button-group`

ricardoarasilva:
> This PR should be divided into 2.
> 
> * The first one to the refactor of `kv-radio-button` to `kv-radio`;
> * The new `kv-radio-button-group`

@miguelpinto98 DONE! I create a new branch called kv-radio with the refactor and the branch kv-radio-button-group is now rebased with kv-radio

ricardoarasilva:
> Looking to `RadioButtonGroup` I'd like to suggest a different approach to identify the active buttons.
> The active buttons should be passed by a dictionary of `{ button-id: state -> (true|false) }`.
> With this we avoid regular iterations over the array, to identify the right toggle and change the value of active state.

@miguelpinto98 Done!

packages/ui-components/src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+	/** When the radio button selection changes, emit the requested tab's key */
+	@Event() radioButtonChange: EventEmitter<string>;
+
+	/** Listen to custom DOM event of radio button selection */
+	@Listen('checkedChange')
+	radioChangeHandler(event: CustomEvent<string>) {
+		this.buttons[event.detail].active = !this.buttons[event.detail].active;
+		this.radioButtonChange.emit(event.detail);
+		console.log(this.buttons);
```
GoncaloCamaz, line 21: Remove console.log

packages/ui-components/src/components/radio-button/radio-button.scss
```diff
@@ -0,0 +1,40 @@
...
+
+	height: var(--button-height);
+	padding: 0 $spacing-3x;
+
+	&.checked {
+		border: 1px solid kv-color('neutral-0');
+	}
+
+	&.disabled {
+		color: kv-color('neutral-4');
```
miguelpinto98, line 38: when disabled, the cursor should change to `default` or `not-allowed`

packages/ui-components/src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -0,0 +1,32 @@
...
+	/** (required) List of radio buttons */
+	@Prop({ reflect: true }) buttons: { [key: string]: IRadioButton } = {};
+
+	/** When the radio button selection changes, emit the requested tab's key */
+	@Event() radioButtonChange: EventEmitter<string>;
+
+	/** Listen to custom DOM event of radio button selection */
+	@Listen('checkedChange')
+	radioChangeHandler(event: CustomEvent<string>) {
+		this.buttons[event.detail].active = !this.buttons[event.detail].active;
```
miguelpinto98, line 19: If we change the `active` value here, we need to do in `kv-radio-button` the same change?

miguelpinto98 (changes requested):
🚀  Good job here, only 2 questions and a rebase to change the status of this PR.

GoncaloCamaz (approved):

packages/ui-components/src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -0,0 +1,32 @@
...
+	tag: 'kv-radio-button-group',
+	styleUrl: 'radio-button-group.scss',
+	shadow: true
+})
+export class KvRadioButtonGroup {
+	/** (required) List of radio buttons */
+	@Prop({ reflect: true }) buttons: { [key: string]: IRadioButton } = {};
+
+	/** When the radio button selection changes, emit the requested tab's key */
+	@Event() radioButtonChange: EventEmitter<string>;
```
djcouto, line 14: This event should emit all radio buttons selected, right? Not only that changed.

djcouto (changes requested):
Left you some comments, mostly about styling. 

Great work 👍

packages/ui-components/src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { IRadioButton } from '../radio-button/radio-button.types';
+
+@Component({
+	tag: 'kv-radio-button-group',
+	styleUrl: 'radio-button-group.scss',
+	shadow: true
+})
+export class KvRadioButtonGroup {
+	/** (required) List of radio buttons */
+	@Prop({ reflect: true }) buttons: { [key: string]: IRadioButton } = {};
```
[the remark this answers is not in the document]
ricardoarasilva: Not really, in this case send the dictionary with the button id and active true in component creation.

packages/ui-components/src/components/radio-button-group/radio-button-group.tsx
ricardoarasilva: This event is emitted every time that a button is checked and return the id of button checked

packages/ui-components/src/components/radio-button-group/radio-button-group.tsx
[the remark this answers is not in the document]
djcouto: Ok, it's different from what I was expected but you're right 👍

packages/ui-components/src/components/radio-button-group/radio-button-group.tsx
djcouto: I got it, I was just saying that since it's a group, the state change should be regarding all buttons instead of only one.

djcouto (approved):
NICE 👍

miguelpinto98 (approved):

miguelpinto98 (changes requested):
@ricardoarasilva I'd notice when we hover the element we don't have any animation, can you define something or talk with @design-team for that.

Other thing: When we active an element we suffer a button blink, because the width increase due to font-weight option. Can we change this behaviour? [Look at here](https://www.sitepoint.com/quick-tip-fixing-font-weight-problem-hover-states/)

packages/ui-components/src/components/radio-button/radio-button.scss
```diff
@@ -0,0 +1,68 @@
...
+	align-items: center;
+
+	border-radius: 2px;
+
+	color: var(--text-color-default);
+
+	background: var(--background-color-default);
+	background-size: 350% 100%;
+	background-position: right bottom;
+	transition: background-position 0.5s ease-out;
```
miguelpinto98, line 48: this animation is working?

miguelpinto98 (approved):

PR #72, opened 2022-04-04
feat: toaster component

GoncaloCamaz:
Please check:
-  Angular and React controllers added
- Should close state be inside toaster component with a prop watch to update the internal state

### DIFF
--- a/packages/angular-ui-components/src/controllers/toaster/toaster-controller.component.ts
+++ b/packages/angular-ui-components/src/controllers/toaster/toaster-controller.component.ts
@@ -1,0 +1,73 @@
+import { Component, OnDestroy, OnInit } from '@angular/core';
+import { Subject } from 'rxjs';
+import { takeUntil, tap } from 'rxjs/operators';
+import { ToasterService } from './toaster-controller.service';
+import { IToaster, EToasterType } from '@kelvininc/ui-components';
+
+@Component({ 
+	selector: 'kv-toaster-controller',
+	template: `
+		<div *ngIf="this.header && this.type">
+			<kv-toaster
+				[header]="header"
+				[description]="description"
+				[ttl]="ttl"
+				[type]="type"
+				(ttlExpired)="onCloseToaster"
+				(clickCloseButton)="onCloseToaster"
+			></kv-toaster>
+		</div>
+	`
+})
+export class ToasterController implements OnInit, OnDestroy {
+	header!: string | null;
+	description?: string = '';
+	type!: EToasterType | null;
+	ttl?: number = 0;
+
+	unsubscriber$: Subject<void> = new Subject();
+
+	constructor(public toastService: ToasterService) { }
+
+	ngOnInit() {
+		this.toastService.getMessageObservable()
+			.pipe(
+				takeUntil(this.unsubscriber$)
+			)
+			.subscribe((message: IToaster) => this.initToaster(message));
+
+		this.toastService.getCloseObservable()
+			.pipe(
+				tap(() => this.onClose()),
+				takeUntil(this.unsubscriber$)
+			)
+			.subscribe();
+	}
+
+	ngOnDestroy() {
+		this.unsubscriber$.next();
+		this.unsubscriber$.complete();
+	}
+
+	onCloseToaster(event: CustomEvent) {
+		event.preventDefault();
+		this.onClose();
+	}
+
+	initToaster(newMessage: IToaster) {
+		this.header = newMessage.header;
+		this.description = newMessage.description ?? '';
+		this.type = newMessage.type;
+
+		if (newMessage && newMessage.ttl) {
+			this.ttl = newMessage.ttl;
+		}
+	}
+
+	onClose() {
+		this.header = null;
+		this.description = '';
+		this.type = null;
+		this.ttl = 0;
+	}
+}
--- a/packages/ui-components/src/components/toaster/toaster.scss
+++ b/packages/ui-components/src/components/toaster/toaster.scss
@@ -1,0 +1,202 @@
+@import '../../assets/styles/globals';
+
+:host {
+	/**
+	 * @prop --toaster-large-height: Toaster with two messages.
+	 * @prop --toaster-small-height: Toaster with one message.
+
+	 * @prop --toaster-width: Toaster width.
+	 * @prop --background-color-default: Toaster background color.
+	 * @prop --kv-color-neutral-2: Main message color.
+	 * @prop --kv-color-neutral-4: Secondary message color.
+	 * @prop --kv-color-neutral-6: Toaster border color.
+
+	 * @prop --info-icon-color: Info color.
+	 * @prop --warning-icon-color: Warning color.
+	 * @prop --error-icon-color: Error color.
+	 * @prop --success-icon-color: Success color.
+
+	 * @prop --reset-button-focused: Close button focused.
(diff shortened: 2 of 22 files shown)


GoncaloCamaz:
Screenshot 2022-04-04 at 10 23 40
Screenshot 2022-04-04 at 10 23 29
Screenshot 2022-04-04 at 10 23 12
Screenshot 2022-04-04 at 10 22 40

GoncaloCamaz:
I removed from previous commits the internal state to close or open the toaster from the component. The reason being having the outside controllers to open/close the toaster. However, I might add an internal state again to close the toaster if the timeout is defined.

packages/ui-components/src/components/toaster/toaster.tsx
```diff
@@ -0,0 +1,72 @@
...
+	private iconTypeClass: EToasterIconTypeClass = TYPE_ICONS[this.type];
+
+	private onCloseClick = event => {
+		window.clearTimeout(this.timeoutID);
+		this.closeToaster();
+		this.clickCloseButton.emit(event);
+	};
+
+	private closeToaster = () => {
+		if (this.ttl > 0) window.clearTimeout(this.timeoutID);
```
GoncaloCamaz, line 33: Something is missing on this function in order to close the toaster. However, I have doubts about its implementation. Since we are using controllers for this component, should the close toaster due to ttl logic be implemented on the controller side or on the toaster side?

packages/angular-ui-components/src/ToasterController.ts
```diff
@@ -0,0 +1,43 @@
+import { Injectable } from '@angular/core';
```
miguelpinto98, line 1: Probably create a separated directory, called `services` and put this file on it. Suggest to rename the file to `ToasterService.ts` or `ToasterController.service.ts`.

Create an index.ts file and add the service to him. After that in public-api export this index instead of this service.

packages/react-ui-components/src/components/ToasterController.types.ts
```diff
@@ -0,0 +1,14 @@
+import { EToasterType } from '@kelvininc/ui-components';
+
+export interface IToasterState {
```
miguelpinto98, line 3: Should this interface extend the interface `IToaster` from Toaster?
Or have separated states, if we use functional components. one to show and another with the toaster info. @djcouto help us here

miguelpinto98 (changes requested):
Good job. Only small questions and improvements. 💪

packages/ui-components/src/components/toaster/toaster.tsx
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	disconnectedCallback() {
+		this.closeToaster();
+	}
+
+	render() {
+		return (
+			<Host>
+				<div
```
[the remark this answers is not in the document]
GoncaloCamaz: I have a separate branch where I implemented css animations to enter and leave (toasterCSSAnimations). I can add those changes to this branch as well. Just need to update the state management.

djcouto (changes requested):
Left my first comments with some suggestions.

Keep going with the great work 💪

packages/ui-components/src/components/toaster/toaster.tsx
```diff
@@ -0,0 +1,72 @@
...
+					class={{
+						'toaster-container': true,
+						[`toaster-container--large`]: !isEmpty(this.description)
+					}}
+				>
+					<div
+						class={{
+							'toaster-icon': true,
+							[`toaster-icon--${this.iconTypeClass.class}`]: true
```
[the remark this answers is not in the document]
GoncaloCamaz: Yep! smart change :D

packages/ui-components/src/components/toaster/toaster.scss
```diff
@@ -0,0 +1,125 @@
...
+	--warning-icon-color: #{kv-color('warning')};
+	--error-icon-color: #{kv-color('error')};
+	--success-icon-color: #{kv-color('success')};
+
+	--close-button-focused: #{kv-color('neutral-2')};
+}
+
+kv-icon {
+	--icon-width: 24px;
+	--icon-height: 24px;
```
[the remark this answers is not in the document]
GoncaloCamaz: I believe they shouldn't change. I will change these properties to avoid being configurable.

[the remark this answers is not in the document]
GoncaloCamaz: These properties are used to define the size of the kv-icon inside the toaster. I did this way to set the size of the icon.

packages/react-ui-components/src/components/ToasterController.types.ts
GoncaloCamaz: I agree with the change.

packages/ui-components/src/components/toaster/toaster.tsx
GoncaloCamaz: Will do some experiments by adding an event to tell the parent component to close the toaster.

packages/angular-ui-components/src/ToasterController.ts
```diff
@@ -0,0 +1,43 @@
...
+
+	success(header: string, description?: string, ttl?: number) {
+		this.publishMessage(EToasterType.Success, header, description, ttl);
+	}
+
+	close() {
+		this.close$.next();
+	}
+
+	getMessageObservable(): Observable<IToaster> {
```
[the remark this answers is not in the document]
miguelpinto98: This function returns an observable, when I think in a stream I have the possibility to make changes on it, but with the observable it isn't possible to make it.

packages/angular-ui-components/src/controllers/toaster-controller.service.ts
```diff
@@ -0,0 +1,43 @@
...
+	close() {
+		this.close$.next();
+	}
+
+	getMessageObservable(): Observable<IToaster> {
+		return this.toasterSubject.asObservable();
+	}
+
+	getCloseObservable(): Observable<void> {
+		return this.close$;
```
miguelpinto98, line 37: `this.close$. asObservable();`

miguelpinto98 (approved):
👍  approved with 1 note

packages/angular-ui-components/src/controllers/toaster/toaster-controller.component.ts
```diff
@@ -0,0 +1,78 @@
...
+@Component({ 
+	selector: 'kv-toaster-controller',
+	template: `
+		<div *ngIf="this.header && this.type">
+			<kv-toaster
+				header="header"
+				description="description"
+				ttl="ttl"
+				type="type"
+				(ttlExpired)="this.onTTLExpired"
```
miguelpinto98, line 16: remove the `this` here

miguelpinto98: we don't need to enable bind in the props? like `[header]=...`,`[description]=...`

miguelpinto98: `ttlExpired` and `onClickCloseButton` can call the same function, since both do the same job

miguelpinto98 (changes requested):
last effort before the approve 😄

miguelpinto98 (approved):
💪

djcouto (approved):
🚀

PR #77, opened 2022-04-13
fix(radio-buttons): change styling accorting to design system

djcouto:


packages/ui-components/src/components/radio-button/radio-button.scss
```diff
@@ -40,29 +40,26 @@
...
-
-	border: 1px solid var(--border-color-default);
 
 	height: var(--button-height);
 	padding: 0 $spacing-3x;
 
 	&.checked {
-		text-shadow: 0 0 0.65px var(--text-color-active), 0 0 0.65px var(--text-color-active);
-		border: 1px solid var(--border-color-active);
+		@include kv-font-label-small-semibold();
```
ricardoarasilva, line 51: When we active an element we suffer a button blink, because the width increase due to font-weight option. Can we change this behaviour? [Look at here](https://www.sitepoint.com/quick-tip-fixing-font-weight-problem-hover-states/)

ricardoarasilva (changes requested):

packages/ui-components/src/components/radio-button/radio-button.scss
djcouto: I didn't know that, it's a nice trick 😅 
However, the result is not quite the same. This gives a blur effect to the font that is not ideal.
Nevertheless, I will rever the shadow at least for know 👍

miguelpinto98 (approved):

ricardoarasilva (approved):

PR #78, opened 2022-04-14
feat(radio-buttons): add manual state control

djcouto:


packages/ui-components/src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -1,30 +1,34 @@
...
 			<Host>
-				{Object.keys(this.buttons).map(key => (
-					<kv-radio-button label={this.buttons[key].label} id={key} disabled={this.buttons[key].disabled} checked={this.buttons[key].active}></kv-radio-button>
+				{this.buttons.map(button => (
+					<kv-radio-button
+						exportparts="radio-button"
+						value={button.value}
+						label={button.label}
+						disabled={this.disabled || button.disabled}
+						checked={button.checked ?? this.selectedButtons.includes(button.value ?? button.label)}
```
miguelpinto98, line 30: make sense to mantain the checked state inside the button configuration? If `selectedButtons` is a dictionary we don't need to make this loop to assert is checked. What you think?

djcouto: Yes, makes sense, I'll change it.

miguelpinto98 (approved):
🚀

pedrorebelocosta (approved):

PR #84, opened 2022-05-02
fix(breadcrumb): change breadcrumb style

djcouto:


packages/ui-components/src/components/breadcrumb-item/breadcrumb-item.types.ts
```diff
@@ -1,6 +1,16 @@
+import { EventEmitter } from '@stencil/core';
 import { IAnchor } from '../../utils/types';
 
 export interface IBreadcrumbItem extends IAnchor {
+	/** (required) The text to display on the breadcrumb */
 	label: string;
+	/** (optional) Sets this breadcrumb styling to be the active one (usually the last one) */
 	active?: boolean;
+	/** (optional) Disabled this breadcrumb touch events (usually the last one) */
+	disabled?: boolean;
```
miguelpinto98, line 10: What is the real difference between active and disabled? Seems the inactive state is the disable state.

packages/ui-components/src/components/breadcrumb-list/breadcrumb-list.scss
```diff
@@ -1,5 +1,29 @@
...
+
+		&::after {
+			@include kv-font-span-regular();
+			display: inline-block;
+			content: var(--breadcrumb-separator-content);
+			margin: 0 $spacing-2x;
+			color: var(--breadcrumb-separator-color);
+		}
+	}
 }
```
miguelpinto98, line 29: 👌

packages/ui-components/src/components/breadcrumb-item/breadcrumb-item.types.ts
djcouto: I'm not completely sure we need this disabled state.

When a breadcrumb item is disabled it means it cannot be clicked.
It's different from inactive because inactive buttons can and should be clickable.

The use case that I found for this is when a breadcrumb item is merely decorative.

But then again, not sure if it's needed.

miguelpinto98 (approved):
🚀

pedrorebelocosta (approved):

PR #87, opened 2022-05-12
feat(tree): add tree component

ricardogoncalves89:
**NIGHT MODE**
>image

**LIGHT MODE**
>image

### DIFF
--- a/packages/ui-components/src/components/tree/tree.types.ts
+++ b/packages/ui-components/src/components/tree/tree.types.ts
@@ -1,0 +1,15 @@
+import { EIconName, EOtherIconName } from '../icon/icon.types';
+import { ETreeItemState } from '../tree-item/tree-item.types';
+
+export interface ITreeNodeItem {
+	id: string; // Should be unique
+	label?: string;
+	additionalLabel?: string;
+	placeholder?: string;
+	icon?: EIconName | EOtherIconName;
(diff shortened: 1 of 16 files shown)


GoncaloCamaz (approved):

pedrorebelocosta (approved):

packages/ui-components/src/components/tree/tree.types.ts
```diff
@@ -0,0 +1,15 @@
+import { EIconName, EOtherIconName } from '../icon/icon.types';
+import { ETreeItemStates } from '../tree-item/tree-item.types';
+
+export interface ITreeNodeItem {
+	id: string; // Should be unique
+	label?: string;
+	additionalLabel?: string;
+	placeholder?: string;
+	icon?: EIconName | EOtherIconName;
```
miguelpinto98, line 9: Should we create a type to icons that combine both enums? Like:

```
export const EIcon = { ...EIconName, ...EOtherIconName };
export type EIcon = typeof EIcon;
```

miguelpinto98 (approved):
🚀 

Approved!
I left a suggestion at the end.

packages/ui-components/src/components/tree/tree.types.ts
ricardogoncalves89: It is better to do it in another pull request because it is used in another places

PR #92, opened 2022-06-17
feat(text-field)!: add min, max and step attributes

djcouto:
This PR introduces a breaking change.
The property for defining the maximum number of characters is now named `maxLength` instead of the previous `max` which now refers to the input maximum value.


miguelpinto98 (approved):
👍

ricardogoncalves89 (approved):

PR #98, opened 2022-07-05
fix(text-field): add numeric input mask

djcouto:


GoncaloCamaz (approved):

packages/ui-components/package.json
```diff
@@ -53,14 +53,16 @@
 	},
 	"dependencies": {
 		"@floating-ui/dom": "^0.4.0",
-		"css-element-queries": "^1.2.3"
+		"css-element-queries": "^1.2.3",
+		"inputmask": "^5.0.8-beta.17"
```
ricardogoncalves89, line 57: Can't you use a release version instead of a beta version?

ricardogoncalves89 (approved):

packages/ui-components/package.json
djcouto: No, because there is an error on the current release version. See this issue for more information https://github.com/RobinHerbots/Inputmask/issues/2581.

PR #108, opened 2022-07-19
fix: move prevent defaults to outside throttler

djcouto:


GoncaloCamaz (approved):

miguelpinto98:
We don't have any test that check if preventDefault is enabled the click shouldn't go forward?

packages/ui-components/src/config.ts
```diff
@@ -0,0 +1,2 @@
+// The number of milliseconds to throttle invocations to.
+export const DEFAULT_THROTTLE_WAIT = 300;
```
miguelpinto98, line 2: 👍

miguelpinto98 (changes requested):
Only to have some responses for my questions @djcouto

packages/ui-components/src/components/radio-button/radio-button.tsx
```diff
@@ -40,13 +41,15 @@ export class KvRadioButton implements IRadioButton, IRadioButtonEvents, IAnchor
...
-		this.clickThrottler = throttle((event: MouseEvent) => {
-			if (this.preventDefault) {
-				event.preventDefault();
-			}
+		this.clickThrottler = throttle(() => this.onCheck(), DEFAULT_THROTTLE_WAIT);
+	}
+
+	onClick(event: MouseEvent) {
+		if (this.preventDefault) {
+			event.preventDefault();
```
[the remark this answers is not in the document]
djcouto: No, we prevent the default behaviour but we still need the `clickThrottler`  to be called.

packages/ui-components/src/components/tree-item/test/tree-item.spec.tsx
```diff
@@ -76,4 +76,31 @@ describe('Tree Item (unit tests)', () => {
 			expect(component.counter).toEqual(32);
 		});
 	});
+
+	describe('when initialized with preventDefault prop', () => {
```
miguelpinto98, line 80: 🔝

PR #113, opened 2022-07-27
feat(modal): add modal component

pedrorebelocosta:
Dark:

image

Light: 

image

### DIFF
--- a/packages/react-ui-components/src/components/ModalOverlay/types.ts
+++ b/packages/react-ui-components/src/components/ModalOverlay/types.ts
@@ -1,0 +1,5 @@
+export interface IModalController {
+	isOpen: boolean;
+	open: () => void;
+	close: () => void;
+}
--- a/packages/ui-components/src/components/modal/modal.base.scss
+++ b/packages/ui-components/src/components/modal/modal.base.scss
@@ -1,0 +1,98 @@
+@import '../../assets/styles/globals';
+
+:host {
+	/**
+	* @prop --modal-height: The modal's height in px.
+	* @prop --modal-width: The modal's width in px.
+	* @prop --modal-header-height: The modal's header height in px.
+	* @prop --modal-footer-height: The modal's footer height in px.
+	* @prop --modal-overlay-z-index: The modal's overlay z-index css property value.
+	* @prop --modal-body-z-index: The modal's body z-index css property value.
+	* @prop --modal-backdrop-color: The modal's backdrop/overlay color.
+	* @prop --modal-background-color: The modal's background color.
+	* @prop --modal-title-text-color: The modal's text header color.
+	* @prop --modal-close-button-width: The modal's close button width
+	* @prop --modal-close-button-height: The modal's close button height
+	*/
+	--modal-height: fit-content;
+	--modal-width: fit-content;
+	--modal-header-height: 60px;
+	--modal-footer-height: unset;
+	--modal-overlay-z-index: 0;
+	--modal-body-z-index: 1;
+	--modal-backdrop-color: #{kv-color('background', 'step-1')};
+	--modal-background-color: #{kv-color('background')};
+	--modal-title-text-color: #{kv-color('kv-text')};
+	--modal-close-button-width: 20px;
+	--modal-close-button-height: 20px;
+}
+
+.modal-overlay {
+	position: fixed;
+	inset: 0;
+	z-index: var(--modal-overlay-z-index);
+	background: var(--modal-backdrop-color);
+}
+
+.modal-container {
+	position: absolute;
+	overflow: auto;
+	outline: none;
+	top: 0;
+	bottom: 0;
+	left: 0;
+	right: 0;
+	margin: auto;
+	width: var(--modal-width);
+	height: var(--modal-height);
+	min-width: 400px;
+	min-height: 235px;
+	border-radius: 10px;
+	background-color: var(--modal-background-color);
+	display: flex;
+	flex-direction: column;
+	z-index: var(--modal-body-z-index);
+
+	.top-bar {
+		display: flex;
+		align-items: center;
+		justify-content: space-between;
+		padding: 0 $spacing-4x;
+		height: var(--modal-header-height);
+
+		.title {
+			@include kv-font-label-small-uppercase-bold;
+			color: var(--modal-title-text-color);
+		}
+
+		.actions {
+			display: inline-flex;
+			align-items: center;
+		}
+	}
+
+	.content {
+		display: flex;
+		flex-direction: column;
+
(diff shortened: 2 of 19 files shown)


packages/react-ui-components/src/components/ModalOverlay/types.ts
```diff
@@ -0,0 +1,8 @@
+import { IModalConfig } from '@kelvininc/ui-components';
+
+export interface IModalController {
+	isOpen: boolean;
+	open: () => void;
+	close: () => void;
+	config?: IModalConfig;
```
pedrorebelocosta, line 7: this is unused, I'll remove

ricardogoncalves89 (approved):
👍

packages/ui-components/src/components/modal/modal.base.scss
```diff
@@ -0,0 +1,93 @@
...
+		flex-direction: column;
+
+		.footer {
+			flex-shrink: 0;
+		}
+	}
+
+	.close {
+		cursor: pointer;
+		width: 28px;
```
djcouto, line 80: Isn't this the `--icon-width`?

packages/ui-components/src/components/modal/modal.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+@Component({
+	tag: 'kv-modal',
+	styleUrls: {
+		night: 'modal.night.scss',
+		light: 'modal.light.scss'
+	},
+	shadow: true
+})
+export class KvModal implements IModalConfig, IModalEvents {
```
djcouto, line 13: Should we add a prop `closable` to define if this modal can be closed by clicking on the close button or by clicking outside of the container?

packages/ui-components/src/components/modal/test/modal.e2e.ts
```diff
@@ -0,0 +1,39 @@
+import { E2EElement, E2EPage, newE2EPage } from '@stencil/core/testing';
+
+describe('Modal (end-to-end)', () => {
+	let page: E2EPage;
+
+	describe('when rendering with default props', () => {
```
djcouto, line 6: Please add tests for when the close button is clicked and when is clicked outside the modal container.

packages/ui-components/src/components/modal/modal.tsx
djcouto: Also, add the click outside logic as we do on the `kv-dropdown` component.

djcouto (changes requested):

packages/ui-components/src/components/modal/modal.tsx
pedrorebelocosta: The modal has the prop 'showCloseButton' which may not be the clearest name, I'll try and change that ;)

When the 'hasBackdrop' flag is enabled, only clicks outside of the modal (as they are adjacent divs) will be dispatched as a close event emitted by `clickClose`. Are you recommending a separation of behaviours or should it be kept this way?

packages/ui-components/src/components/modal/modal.base.scss
pedrorebelocosta: Would adding a css property/variable to the modal where you can customize the width/height of this button and use that to assign a value on the kv-icon width/height property work for flexibility and reusability?

i.e:

```
:host {
	(...)
	--modal-close-button-width: 20px;
	--modal-close-button-height: 20px;
	(...)
}

.close {
	kv-icon {
		--icon-width: var(--modal-close-button-width);
		--icon-height: var(--modal-close-button-height);
	}
}
```

packages/ui-components/src/components/modal/modal.tsx
pedrorebelocosta: I'll be adding the "closable" property to the interface as we discussed and control the emission of close events using that flag, thanks!

packages/ui-components/src/components/modal/test/modal.e2e.ts
pedrorebelocosta: done ;)

packages/ui-components/src/components/modal/modal.tsx
pedrorebelocosta: If there's a preference for the outside click detection like the one on the `kv-dropdown` component please let me know, it's a small change :P

packages/ui-components/src/components/modal/modal.base.scss
```diff
@@ -0,0 +1,97 @@
...
+	--modal-background-color: #{kv-color('background')};
+	--modal-title-text-color: #{kv-color('kv-text')};
+	--modal-close-button-width: 20px;
+	--modal-close-button-height: 20px;
+}
+
+.modal-overlay {
+	position: fixed;
+	inset: 0;
+	z-index: 1000;
```
miguelpinto98, line 32: why use so higher z-index? 
This should use base low number.

packages/ui-components/src/components/modal/modal.tsx
```diff
@@ -0,0 +1,67 @@
...
+	/** @inheritdoc */
+	@Prop() label?: string;
```
miguelpinto98, line 15: Should this be named as `title` ?

miguelpinto98 (approved):
🚀  Good job here. Seems to have a base to iterate to other kinds of modals.
Only check my 2 comments below before merge.

packages/ui-components/src/components/modal/modal.tsx
pedrorebelocosta: Doing that would set off alarms in stencil's linter since it's a reserved keyword in HTML, maybe `headerTitle` or something similar?

packages/ui-components/src/components/modal/modal.base.scss
pedrorebelocosta: If I'm not mistaken I took the z values from Angular's CDK overlay, I agree it could be lower, but how much lower?

pedrorebelocosta: As we discussed, I've created a css variable so we can modify this value externally, current values are 0 for the overlay and 1 for the modal body.

packages/ui-components/src/components/modal/modal.tsx
miguelpinto98: `headerTitle` is good

djcouto (approved):
:shipit:

packages/ui-components/src/components/modal/modal.base.scss
djcouto: CSS properties option seems better to me 👍

PR #115, opened 2022-08-01
feat: toggle tip component

GoncaloCamaz:


GoncaloCamaz:
Screenshot 2022-08-03 at 15 07 35

packages/ui-components/src/components/toggle-tip/test/toggle-tip.e2e.ts
```diff
@@ -0,0 +1,11 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+describe('kv-toggle-tip', () => {
```
ricardogoncalves89, line 3: please add more test scenarios

ricardogoncalves89 (changes requested):
Why you don't add an extra slot to the `KvTooltip` you only need that to have the same result.

GoncaloCamaz:
> Why you don't add an extra slot to the `KvTooltip` you only need that to have the same result.

I used a new component since the events to handle this component are not the same as the ones in tooltip. However, this two components are actually very similar, so, a possible alternative approach could consist on adding a prop like an enum to define the events that should be listened in order to operate the tooltip

packages/ui-components/src/components/toggle-tip/toggle-tip.scss
```diff
@@ -0,0 +1,44 @@
...
+	 * @prop --button-icon-width: Button's icon width.
+	 * @prop --button-icon-height: Button's icon height.
+	 */
+
+	--button-icon-width: 16px;
+	--button-icon-height: 16px;
+}
+
+kv-icon {
+	--icon-width: var(--button-icon-width);
```
djcouto, line 14: No need for this. Export the `kv-icon` part so it can be styled by the toggle parent.

packages/ui-components/src/components/toggle-tip/toggle-tip.types.ts
```diff
@@ -0,0 +1,14 @@
+export enum EToggleTipPosition {
```
djcouto, line 1: Isn't this the same as `ETooltipPosition`?

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * @part toggle-tip-container - The toggle tip container.
+ * @part toggle-tip-slot-content - The toggle tip slot content.
+ */
+@Component({
+	tag: 'kv-toggle-tip',
+	styleUrl: 'toggle-tip.scss',
+	shadow: true
+})
+export class KvToggleTip {
```
djcouto, line 17: Please add the component to the angular module in `kv-ui-components.module` file

djcouto (changes requested):

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
miguelpinto98, line 17: Isn't possible to extend the `KvToggleTip` by `KvTooltip`?
Or create a generic class that have pratically all the code that are shared between the compoents and extend both class that component @GoncaloCamaz ?

packages/react-ui-components/src/stories/Components/ToggleTip.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+		},
+		allowedPositions: { control: { type: 'multi-select' }, options: Object.values(EToggleTipPosition) }
+	},
+	parameters: {
+		notes: require('@ui-notes/toggle-tip/readme.md'),
+		layout: 'centered'
+	}
+};
+
+const TooltipTemplate: ComponentStory<typeof KvToggleTip> = args => (
```
pedrorebelocosta, line 28: `ToggletipTemplate`

packages/ui-components/src/components/toggle-tip/test/toggle-tip.e2e.ts
pedrorebelocosta, line 3: please add tests

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
```diff
@@ -0,0 +1,87 @@
...
+				})
+			);
+		}
+
+		const onButtonClick = () => {
+			this.toggleTipOpen = !this.toggleTipOpen;
+			if (this.toggleTipOpen) {
+				toggletip.style.display = 'inline-block';
+				this.updatePosition(toggleIcon, toggletip, position, middleware);
+			} else toggletip.style.display = '';
```
pedrorebelocosta, line 56: we don't tipically use single line `if/else` statements, please add brackets to the `else` block

packages/ui-components/src/components/toggle-tip/toggle-tip.types.ts
pedrorebelocosta: Since we're using the FloatingUI lib to position our elements relatively could we define this enum in the `packages/ui-components/src/utils/types/components.ts` types file?

pedrorebelocosta (changes requested):
A few minor changes

packages/ui-components/src/components/toggle-tip/toggle-tip.scss
```diff
@@ -0,0 +1,45 @@
...
+	word-wrap: break-word;
+	color: kv-color('neutral-0');
+	background-color: kv-color('neutral-6');
+	box-shadow: 0 8px 16px -4px kv-color('neutral-6', 'base', 0.25), 0 0 1px 0 kv-color('neutral-3', 'base', 0.31);
+	user-select: none;
+	border-radius: 2px;
+	padding: $spacing-3x;
+	position: fixed;
+	display: none;
+	z-index: 10000;
```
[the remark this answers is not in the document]
GoncaloCamaz: @miguelpinto98  was just a test to check if it could fix the position under the cards but i believe there is another solution for that problem. The actual problem relates to the size of the div where the icon to open the toggle is placed. If the div is too small, we will have problems with the toggle and with the tooltip component. Therefore i will remove this z-index since it was not the solution for the problem.

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
GoncaloCamaz: @miguelpinto98 i added now some code in order to check for clicks outside of the toggle tip. Therefore and since the behaviours of this two components are distinct regarding opening and closing, i believe its better to separate both but please feel free to suggest otherwise. 
I will try to extend the tooltip props for the toggle tip.

packages/ui-components/src/components/toggle-tip/toggle-tip.scss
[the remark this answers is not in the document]
GoncaloCamaz: @miguelpinto98 replaced by z-index: 1000; since in the case of the charts from the highcharts, the toggle tip was being drawn behind the chart.

packages/ui-components/src/components/toggle-tip/readme.md
```diff
@@ -0,0 +1,50 @@
+# kv-toggletip
```
djcouto, line 1:
```suggestion
# _kv-toggle-tip_
```

djcouto (approved):
LGMT 👍

Insatisfeito (approved):

packages/ui-components/src/components/toggle-tip/toggle-tip.scss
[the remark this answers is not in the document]
pedrorebelocosta: suggestion: define a css prop for it, in our use case with highcharts you'll need this higher z-index, however other library users may not need such a high z-index.

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
```diff
@@ -0,0 +1,105 @@
...
+ * @part toggle-tip-slot-content - The toggle tip slot content.
+ */
+@Component({
+	tag: 'kv-toggle-tip',
+	styleUrl: 'toggle-tip.scss',
+	shadow: true
+})
+export class KvToggleTip implements IToggleTip, IToggleTipEvents {
+	/** @inheritdoc */
+	@Prop({ reflect: true }) text!: string;
```
ricardogoncalves89, line 20: If I only need to use the slot to show the necessary info, I can't because the text is mandatory

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
```diff
@@ -0,0 +1,105 @@
...
+		};
+		toggleIcon.addEventListener('mousedown', onButtonClick, { once: true });
+	};
+
+	render() {
+		return (
+			<Host>
+				<kv-icon id="toggle-tip-content" name={this.icon} exportparts="icon" />
+				<div id="toggle-tip" class="toggle-tip-container" part="toggle-tip-container">
+					<div class="toggle-tip-text">{this.text}</div>
```
ricardogoncalves89, line 97: this should be displayed only if `text` is defined

packages/ui-components/src/components/toggle-tip/usage/react.md
```diff
@@ -0,0 +1,24 @@
+```tsx
+import React from 'react';
+
+import { KvToggleTip, KvActionButton, ETooltipPosition, EActionButtonType } from '@kelvininc/react-ui-components';
+
+export const TagLetterExample: React.FC = () => (
```
ricardogoncalves89, line 6: ToggleTooltipExample

ricardogoncalves89 (dismissed):

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
GoncaloCamaz: <div class="toggle-tip-text">{this.text}</div> this here correct? i will remove the ! from text and place here a verification

miguelpinto98:
Screenshot 2022-08-17 at 12 23 14

For this behaviour I think this component can't expose a tip for a undetermined component. 
So I think is necessary to remove the usage of kv-icon and add a slot. Or maybe if we don't have a slot we put the kv-icon.

pedrorebelocosta (approved):

packages/react-ui-components/src/stories/Components/ToggleTip.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+	},
+	parameters: {
+		notes: require('@ui-notes/toggle-tip/readme.md'),
+		layout: 'centered'
+	}
+};
+
+const ToggleTipTemplate: ComponentStory<typeof KvToggleTip> = args => (
+	<KvToggleTip {...args}>
+		<KvIcon slot="opening-element-slot" name={EIconName.Info} />
```
miguelpinto98, line 33: rename ``slot="open-element-slot"``

packages/react-ui-components/src/stories/Components/ToggleTip.stories.tsx
```diff
@@ -0,0 +1,55 @@
+import { ComponentStory } from '@storybook/react';
+import React from 'react';
+import { EActionButtonType, KvActionButton, ETooltipPosition, KvToggleTip, EIconName, EOtherIconName, KvIcon } from '../../components';
+
+KvToggleTip.displayName = 'KvToggleTip';
+
+export default {
+	title: 'Feedback/ToggleTip',
+	component: 'kv-toggle-tip',
+	argTypes: {
```
miguelpinto98, line 10: add isOpen Property

packages/ui-components/src/components/select-option/select-option.scss
```diff
@@ -37,7 +37,8 @@
 	.text-container {
 		display: flex;
 		flex-direction: column;
-
+		min-width: 0;
```
miguelpinto98, line 40: thanks 🙏

packages/ui-components/src/components/toggle-tip/toggle-tip.scss
```diff
@@ -0,0 +1,37 @@
+@import '../../assets/styles/globals';
+
+:host {
+	/**
+	 * @prop --toggletip-z-index: toggle tip z-index.
+	 */
+
+	--toggletip-z-index: 1000;
```
miguelpinto98, line 8: 1 isn't enough?

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
```diff
@@ -0,0 +1,104 @@
...
+	/** @inheritdoc */
+	@Prop({ reflect: true }) text?: string;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) position?: ETooltipPosition;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) allowedPositions?: ETooltipPosition[];
+	/** @inheritdoc */
+	@Prop({ reflect: true }) options?: Partial<ComputePositionConfig>;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) isOpen?: boolean = false;
```
miguelpinto98, line 27: don't forget to add a property that check on mount if this prop is true to open the toggleTIp

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
```diff
@@ -0,0 +1,104 @@
...
+					...DEFAULT_AUTOPLACEMENT_CONFIG,
+					allowedPlacements: this.allowedPositions
+				})
+			);
+		}
+
+		const onButtonClick = () => {
+			this.isOpen = !this.isOpen;
+			if (this.isOpen) {
+				toggletip.style.display = 'inline-block';
```
miguelpinto98, line 70: we always need to control the tooltip like this?

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
```diff
@@ -0,0 +1,104 @@
...
+		if (isEmpty(position)) {
+			middleware.push(
+				autoPlacement({
+					...DEFAULT_AUTOPLACEMENT_CONFIG,
+					allowedPlacements: this.allowedPositions
+				})
+			);
+		}
+
+		const onButtonClick = () => {
```
miguelpinto98, line 67: I was wondering if it's possible to move this function to out of this scope, but seems difficult. Only if we bind the callback with arguments that we need to run the floating/ui

miguelpinto98 (changes requested):
Almost almost!!! 💪

packages/react-ui-components/src/stories/Components/ToggleTip.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+		text: { type: 'text' },
+		position: {
+			control: { type: 'select' },
+			options: Object.values(ETooltipPosition)
+		},
+		icon: {
```
djcouto, line 16: The icon is no longer on this component. Please remove this

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
GoncaloCamaz: Now we do but I added an internal state to enable opening from the outside

packages/ui-components/src/components/toggle-tip/toggle-tip.scss
```diff
@@ -0,0 +1,48 @@
...
+	--toggle-tip-container-max-width: 292px;
+}
+
+.toggle-tip-open-element-wrapper {
+	width: fit-content;
+	height: fit-content;
+}
+
+.toggle-tip-container {
+	@include kv-font-span-semibold();
```
miguelpinto98, line 21: we need this include here, since it is defined on `.toggle-tip-text` too?

miguelpinto98 (changes requested):
After this recent changes, i saw somethings that can be improved in my perspective. Please, check it. 💪

packages/ui-components/src/components/toggle-tip/toggle-tip.types.ts
```diff
@@ -0,0 +1,27 @@
...
+	allowedPositions?: ETooltipPosition[];
+	/** (optional) Object with toggle tip options */
+	options?: Partial<ComputePositionConfig>;
+	/** (optional) Set open state of toggle tip, default false */
+	isOpen?: boolean;
+	/** (optional) if true it will ignore outside clicks to close the toggle tip */
+	fixedTip?: boolean;
+	/** (optional) if true it will render an arrow pointing to the opening element */
+	arrow?: boolean;
+	/** (optional) if true it will clicks to open toggle tip */
```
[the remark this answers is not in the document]
GoncaloCamaz: a "disable"is missing, will fix

miguelpinto98 (approved):
🚀  💪

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
```diff
@@ -0,0 +1,150 @@
...
+			}
+		}
+	}
+
+	componentDidRender() {
+		this.calculateTogglePositions();
+	}
+
+	private getToggleOpenElement = (): HTMLElement | null => {
+		return this.el.shadowRoot.querySelector('#toggle-tip-open-element-wrapper') as HTMLElement | null;
```
djcouto, line 61: The open element, tip container element and arrow element should be passed as prop and fallback to this selector.

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
```diff
@@ -0,0 +1,150 @@
...
+})
+export class KvToggleTip implements IToggleTip, IToggleTipEvents {
+	/** @inheritdoc */
+	@Prop({ reflect: true }) text?: string;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) position?: ETooltipPosition;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) allowedPositions?: ETooltipPosition[];
+	/** @inheritdoc */
+	@Prop({ reflect: true }) options?: Partial<ComputePositionConfig>;
```
djcouto, line 26: This is not being used.

djcouto (changes requested):
Nice work 👍

Left a few comments and suggestions.

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
GoncaloCamaz: Yep, this was used on the  autoUpdate function but forgot to remove.

djcouto: Not sure if you should remove it or use it, but to keep this PR small let's keep it this way.

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
```diff
@@ -0,0 +1,150 @@
...
+	};
+
+	private addButtonClickEvent = () => {
+		const toggleIcon = this.getToggleOpenElement();
+
+		const onButtonClick = () => {
+			this.isOpen = !this.isOpen;
+			this.openStateChange.emit(this.isOpen);
+		};
+		toggleIcon.addEventListener('mousedown', onButtonClick, { once: true });
```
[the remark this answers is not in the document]
GoncaloCamaz: Without once true, toggle tip needs double click to open

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
```diff
@@ -0,0 +1,150 @@
...
+			Object.assign(toggletip.style, {
+				left: `${x}px`,
+				top: `${y}px`
+			});
+
+			if (this.withArrow === true) {
+				const { x: arrowX, y: arrowY } = middlewareData.arrow;
+				Object.assign(arrowElement.style, {
+					...getArrowElementPositionConfig(arrowX, arrowY, placement)
+				});
```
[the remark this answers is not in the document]
GoncaloCamaz: Cannot assign read only :/

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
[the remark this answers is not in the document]
pedrorebelocosta: This may do for now as a "hacky" workaround, however I suspect this could have something to do with event propagation.

As per the [docs](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#once) `once` removes the event listener after the first event fires so I'm not sure if this solution is sustainable in the long run.

pedrorebelocosta (approved):

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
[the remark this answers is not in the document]
GoncaloCamaz: @pedrorebelocosta thanks 🙇

packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
djcouto: Don't forget this.

djcouto (approved):

PR #116, opened 2022-08-03
fix(kv-dropdown): open dropdown list at the top

ricardogoncalves89:
Open dropdown list at the top if  there is no space available at the bottom

Captura de ecrã 2022-08-03, às 12 23 56
Captura de ecrã 2022-08-03, às 12 24 15

### DIFF
--- a/packages/ui-components/src/components/dropdown/dropdown.tsx
+++ b/packages/ui-components/src/components/dropdown/dropdown.tsx
@@ -1,3 +1,4 @@
+import { autoUpdate, computePosition, flip, offset } from '@floating-ui/dom';
 import { Component, Host, h, Prop, Event, EventEmitter, Listen, Element } from '@stencil/core';
 import { EIconName, EOtherIconName } from '../icon/icon.types';
 import { EInputFieldType, EValidationState } from '../text-field/text-field.types';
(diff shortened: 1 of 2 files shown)


djcouto (approved):
🚀

packages/ui-components/src/components/dropdown/dropdown.tsx
```diff
@@ -51,11 +52,44 @@ export class KvDropdown implements IDropdown, IDropdownEvents {
...
+					placement: 'bottom',
+					middleware: [
+						offset(8),
+						flip({
+							padding: 15,
+							fallbackPlacements: ['top']
+						})
+					]
+				}).then(({ x, y }) => {
+					Object.assign(dropdownList.style, {
```
Insatisfeito, line 74: Could we mutate the object directly here?
dropdownList.style.left = `${x}px`;

Or atleast we should prefer object spread over assign.

ricardogoncalves89: I think so. I used this because it's the suggested in the FloatingUI lib.

Insatisfeito: I would assume it is because of older browser support. It's safer than mutating directly since the object might not exist (?), but slower than the other approaches, it's up yo you if the change makes sense, just commenting so we don't start using `Object.assign` as a rule of thumb.

Insatisfeito (approved):

PR #117, opened 2022-08-03
feat(range): add new range component

GoncaloCamaz:


GoncaloCamaz:
Screenshot 2022-08-03 at 13 50 33

packages/ui-components/src/components/range/range.scss
```diff
@@ -0,0 +1,93 @@
+@import '../../assets/styles/globals';
+
+:host {
+	/** 
+	 * @prop --range-width Width of the range slider
+	*/
+
+	--range-width: 600px;
```
ricardogoncalves89, line 8: the width should be fit to the parent container or not?

packages/ui-components/src/components/range/range.tsx
```diff
@@ -0,0 +1,92 @@
+import { Component, Host, h, Prop, Element, EventEmitter, Event, State, Watch } from '@stencil/core';
+import { IRange, IRangeEvents } from './range.types';
+
+@Component({
+	tag: 'kv-range',
+	styleUrl: 'range.scss',
+	shadow: true
+})
+export class KvRange implements IRange, IRangeEvents {
```
ricardogoncalves89, line 9: Please add the component to the angular module in `kv-ui-components.module` file

packages/ui-components/src/components/range/range.tsx
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+		slider.oninput = () => {
+			this.applyCssStyles(true);
+		};
+	}
+
+	private applyCssStyles = (launchValueChangeEvent: boolean) => {
+		const rangeInputValue = this.el.shadowRoot.querySelector('input');
+		const inputElement = this.el.shadowRoot.querySelector('input') as HTMLElement;
```
ricardogoncalves89, line 47: these two consts are the same  `rangeInputValue` `inputElement`

packages/ui-components/src/components/range/test/range.e2e.ts
```diff
@@ -0,0 +1,11 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+describe('kv-range', () => {
```
ricardogoncalves89, line 3: please add more test scenarios

packages/ui-components/src/components/range/test/range.spec.tsx
```diff
@@ -0,0 +1,19 @@
+import { newSpecPage, SpecPage } from '@stencil/core/testing';
+import { KvRange } from '../range';
+
+describe('Range (unit tests)', () => {
+	let page: SpecPage;
+
+	describe('when uses default props', () => {
```
ricardogoncalves89, line 7: please add more test scenarios

packages/ui-components/src/components/range/range.scss
```diff
@@ -0,0 +1,93 @@
+@import '../../assets/styles/globals';
+
+:host {
+	/** 
+	 * @prop --range-width Width of the range slider
```
djcouto, line 5: missing the dots after `--range-width`

packages/ui-components/src/components/range/range.scss
```diff
@@ -0,0 +1,93 @@
...
+	border-radius: $spacing;
+	background: kv-color('neutral-9');
+
+	&::-webkit-slider-thumb {
+		width: 48px;
+		height: 48px;
+		cursor: pointer;
+		z-index: 3;
+		position: relative;
+		/* stylelint-disable */
```
djcouto, line 39: Please add some justification or links to why you need to disable this. Just for future reference.

packages/ui-components/src/components/range/range.types.ts
```diff
@@ -0,0 +1,17 @@
+import { EventEmitter } from '@stencil/core';
+
+export interface IRangeEvents {
```
djcouto, line 3: export this file on `packages/ui-components/src/types.ts`

packages/ui-components/src/components/range/range.tsx
```diff
@@ -0,0 +1,92 @@
...
+		const selector = this.el.shadowRoot.getElementById('selector');
+
+		let inputValue = rangeInputValue.value;
+		const percentage = this.getInputPercentageFromValue(inputValue);
+		const offSet = this.getOffset(percentage);
+
+		selector.style.left = percentage + '%';
+		selector.style.marginLeft = offSet + 'px';
+		inputElement.style.background = `linear-gradient(90deg, #005cc7 ${percentage}%, #121212 ${percentage + 0.1}%)`;
```
djcouto, line 56: Hum, probably this could be moved to a `getInputBackground(percentage: number): string` helper

packages/ui-components/src/components/range/range.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+		if (launchValueChangeEvent) this.onValueChange(parseInt(inputValue));
+	};
+
+	private getInputPercentageFromValue = (inputValue: string) => {
+		return (100 * (parseInt(inputValue) - this.min)) / (this.max - this.min);
+	};
+
+	private getOffset = (percentage: number) => {
+		const invertPercentage = 1 - percentage / 100;
```
djcouto, line 66: Should this be a helper?

djcouto (changes requested):

packages/ui-components/src/components/range/range.scss
```diff
@@ -0,0 +1,93 @@
...
+	border: $spacing solid kv-color('neutral-0');
+	top: 25px;
+	height: 18px;
+	width: 18px;
+	border-radius: 50%;
+	background: #005cc7;
+	cursor: pointer;
+	/* stylelint-disable */
+	-webkit-appearance: none;
+	/* stylelint-enable */}
```
pedrorebelocosta, line 69: \n here

pedrorebelocosta (changes requested):

packages/ui-components/src/components/range/range.tsx
ricardogoncalves89: and you have another problem you can't have fixed colors that way the component  will never adapt to the theme

packages/ui-components/src/components/range/range.scss
GoncaloCamaz: @djcouto  i disabled here the lint because i was using -webkit-appearence: none and this property is deprecated. Now i replaced with appearence: none and I dont need to disable lint anymore so it will be removed.

packages/ui-components/src/components/range/range.tsx
GoncaloCamaz: @ricardogoncalves89 could we consider receiving this two colors as props? Like "backgroundColor: string; sliderColor: string"?

pedrorebelocosta: Suggestion: Use css props to manipulate that from the implementation side, if needed you can then use JS to manipulate the css props as you see fit.

Something to help: https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty

Let me know if you need any help ;)

packages/ui-components/src/components/range/range.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+		const selector = this.el.shadowRoot.getElementById('selector');
+
+		let inputValue = rangeInputValue.value;
+		const percentage = getInputPercentageFromValue(inputValue, this.min, this.max);
+		const offSet = getOffset(percentage);
+
+		selector.style.left = percentage + '%';
+		selector.style.marginLeft = offSet + 'px';
+		rangeInputValue.style.background = `linear-gradient(90deg, #005cc7 ${percentage}%, #121212 ${percentage + 0.1}%)`;
```
ricardogoncalves89, line 42: You can't have hardcoded colors like you have here. 
Suggestion: 
`rangeInputValue.style.background = linear-gradient(90deg, var(--slider-background-filled) ${percentage}%, var(--slider-background-empty) ${percentage + 0.1}%);`
and  add these var in scss file

packages/ui-components/src/components/range/test/range.e2e.ts
```diff
@@ -0,0 +1,34 @@
+import { E2EPage, EventSpy, newE2EPage } from '@stencil/core/testing';
+
+describe('Range (end to end)', () => {
+	let page: E2EPage;
+
+	describe('kv-range', () => {
+		it('renders', async () => {
+			page = await newE2EPage();
```
ricardogoncalves89, line 8: test preparation should be inside the `beforeEach`

packages/ui-components/src/components/range/test/range.e2e.ts
```diff
@@ -0,0 +1,34 @@
+import { E2EPage, EventSpy, newE2EPage } from '@stencil/core/testing';
+
+describe('Range (end to end)', () => {
+	let page: E2EPage;
+
+	describe('kv-range', () => {
```
ricardogoncalves89, line 6: more descriptive here please

packages/ui-components/src/components/range/range.tsx
```diff
@@ -0,0 +1,69 @@
...
+	@Prop({ reflect: true }) value?: number = 0;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) step?: number = 1;
+	/** @inheritdoc */
+	@Event() valueChange: EventEmitter<number>;
+
+	componentDidRender() {
+		const slider = this.el.shadowRoot.getElementById('slider') as HTMLElement;
+		this.applyCssStyles();
+		slider.oninput = () => {
```
ricardogoncalves89, line 26: Can you put these directly in the JSX or not? 
`<input id="slider" class="slider" type="range" min={this.min} max={this.max} value={this.value} step={this.step}  oninput={this.applyCssStyles}/>`

ricardogoncalves89 (changes requested):

packages/ui-components/src/components/range/range.tsx
ricardogoncalves89: answered in last review.

packages/ui-components/src/components/range/range.tsx
GoncaloCamaz: @ricardogoncalves89 updated on new pr

PR #118, opened 2022-08-04
feat: add slots to toaster component

GoncaloCamaz:


packages/ui-components/src/components/toaster/toaster.tsx
```diff
@@ -111,6 +111,7 @@ export class KvToaster implements IToaster, IToasterEvents {
 						<span class="main-message">{this.header}</span>
 						{!isEmpty(this.description) && <span class="secondary-message">{this.description}</span>}
 					</div>
+					<slot></slot>
 					{this.closable && (
 						<div class="toaster-close-icon">
```
djcouto, line 116: Just to prevent future features, I think this slot should be a named slot "right" and we should also have a named left slot too.

packages/react-ui-components/src/stories/Components/Toaster.stories.tsx
```diff
@@ -32,6 +32,11 @@ export default {
 };
 
 const ToasterTemplate: ComponentStory<typeof KvToaster> = args => <KvToaster {...args} />;
+const ToasterWithSlots: ComponentStory<typeof KvToaster> = args => (
```
djcouto, line 35: Missing the `Template` suffix.

djcouto (changes requested):

packages/ui-components/src/components/toaster/toaster.tsx
GoncaloCamaz: @djcouto where would you add the left slot?

djcouto: Before the "message-content" and after the icon.

pedrorebelocosta (approved):

packages/ui-components/src/components/toaster/toaster.tsx
GoncaloCamaz: @djcouto this would introduce some breaking changes to the slot component. Should i proceed with the change anyway?

djcouto: Not sure which breaking changes you are referring to, but let's keep the component without the left slot and implement it in the future when we need it.

djcouto (approved):

PR #123, opened 2022-08-11
fix(dropdown): fix check for click outside fuction

laraujo7:
When clicking outside this dropdown wouldn't close, I was trying to figure out the problem and realized that the console was giving this error.
2022-08-11-164133_503x199_scrot

Digging deeper, I figured out that this error was actually the dropdown problem. So I made some modifications to the function, and it seems to be working. My only doubt  is the type of the target (it was suggested by VSCode).

### DIFF
--- a/packages/ui-components/src/components/dropdown/dropdown.tsx
+++ b/packages/ui-components/src/components/dropdown/dropdown.tsx
@@ -38,11 +38,10 @@ export class KvDropdown implements IDropdown, IDropdownEvents {
 	@Element() el: HTMLKvDropdownElement;
 
 	@Listen('click', { target: 'window' })
-	checkForClickOutside(ev: { path: HTMLElement[] }) {
-		if (ev.path.some(element => element === this.el)) {
+	checkForClickOutside(ev: { target: Node }) {
+		if (this.el.contains(ev.target)) {
 			return;
 		}
-
 		this.isOpen = false;
 		this.openStateChange.emit(this.isOpen);
 	}


nelsonmestevao (approved):

packages/ui-components/src/components/dropdown/dropdown.tsx
```diff
@@ -38,15 +38,13 @@ export class KvDropdown implements IDropdown, IDropdownEvents {
...
-		if (ev.path.some(element => element === this.el)) {
+	checkForClickOutside(ev: { target: Node }) {
+		if (this.el.contains(ev.target)) {
 			return;
 		}
-
 		this.isOpen = false;
 		this.openStateChange.emit(this.isOpen);
 	}
-
```
miguelpinto98, line 49: +

packages/ui-components/src/components/dropdown/dropdown.tsx
```diff
@@ -29,7 +29,7 @@ export class KvDropdown implements IDropdown, IDropdownEvents {
 	@Prop({ reflect: true }) helpText?: string;
 	/** @inheritdoc */
 	@Prop({ reflect: true }) disabled?: boolean;
-	/** @inheritdoc */
+	/** @inheritd oc */
```
miguelpinto98, line 32: this change was unintended?

miguelpinto98 (approved):
👍

djcouto (approved):

miguelpinto98:
@laraujo7 rebase with dev 🙏

PR #129, opened 2022-08-15
feat(select-option): add bottom slot for content

pedrorebelocosta:
image

### DIFF
--- a/packages/react-ui-components/src/stories/Components/SelectOption.stories.tsx
+++ b/packages/react-ui-components/src/stories/Components/SelectOption.stories.tsx
@@ -1,5 +1,5 @@
 import { ComponentStory } from '@storybook/react';
-import React from 'react';
+import React, { CSSProperties } from 'react';
 import { KvSelectOption } from '../../components';
 
 // Required to have the correct TagName in the code sample
@@ -31,6 +31,18 @@ export default {
 };
 
 const SelectOptionTemplate: ComponentStory<typeof KvSelectOption> = args => <KvSelectOption {...args}></KvSelectOption>;
+const BottomSlotSelectOptionTemplate: ComponentStory<typeof KvSelectOption> = args => {
+	const cssProps = {
+		'--select-item-height': 'unset',
+		'--select-item-flex-alignment': 'flex-start'
+	} as CSSProperties;
+
+	return (
+		<KvSelectOption {...args} style={cssProps}>
+			{args.children}
+		</KvSelectOption>
+	);
+};
 
 export const Default = SelectOptionTemplate.bind({});
 Default.args = {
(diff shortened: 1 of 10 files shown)


packages/ui-components/src/components/select-option/select-option.scss
```diff
@@ -2,66 +2,76 @@
...
+		opacity: 0.5;
+		pointer-events: none;
+	}
 
-		kv-icon {
-			display: inline-flex;
-			margin-right: $spacing-2x;
-		}
+	&.has-bottom-slot {
+		height: fit-content;
```
miguelpinto98, line 68: this can be removed I guess. If we want to fit all the content we can set the `--select-item-height` to `unset`

packages/ui-components/src/components/select-option/select-option.scss
```diff
@@ -2,66 +2,76 @@
...
+		padding: $spacing-2x $spacing-3x;
+		align-items: flex-start;
```
miguelpinto98, line 70: probably is better to define this is a variable too

miguelpinto98 (changes requested):
Seems good, only two things to think about.

packages/react-ui-components/src/stories/Components/SelectOption.stories.tsx
```diff
@@ -40,3 +40,10 @@ Default.args = {
 	togglable: true,
 	disabled: false
 };
+
+export const BottomSlot = SelectOptionTemplate.bind({});
+BottomSlot.args = {
+	...Default.args,
+	hasBottomSlot: true,
+	children: <div>Hello</div>
```
djcouto, line 48: 👍

packages/ui-components/src/components/select-option/select-option.scss
```diff
@@ -2,66 +2,76 @@
...
-			background-color: var(--dropdown-item-background-color-selected);
+	&.selected {
+		background-color: var(--select-item-background-color-selected);
 
-			.item-label {
-				@include kv-font-span-semibold;
-			}
+		.item-label {
+			@include kv-font-span-semibold;
+			color: var(--select-item-label-color-selected);
```
djcouto, line 54: 👍

packages/ui-components/src/components/select/select.scss
```diff
@@ -4,16 +4,20 @@
...
+	 * @prop --select-border-width: Select border width.
+	 * @prop --select-border-radius: Select border radius.
 	 * @prop --select-border-color: Select border color.
 	 */
 
 	--select-max-height: auto;
 	--select-background-color: #{kv-color('neutral-7')};
+	--select-border-width: 1px;
+	--select-border-radius: #{$spacing};
 	--select-border-color: #{kv-color('neutral-6')};
```
djcouto, line 16: Nice ideia 👍 

Just a thought two avoid two CSS custom properties for the "same" property:
```suggestion
	--select-border: 1px solid #{kv-color('neutral-6')};
	--select-border-radius: #{$spacing};
```

djcouto (approved):
Well done 💪

Insatisfeito (approved):

ricardogoncalves89 (commented):
Please take a look at the comments

PR #131, opened 2022-08-17
feat(range): add new range component

GoncaloCamaz:


Insatisfeito (approved):

packages/ui-components/src/components/range/range.scss
```diff
@@ -0,0 +1,92 @@
...
+
+:host {
+	/** 
+	 * @prop --range-width: Width of the range slider
+	 * @prop --slider-background-filled: color of the slider when its full
+	 * @prop --slider-background-empty: color of the slider when its empty
+	*/
+
+	--range-width: 100%;
+	--slider-background-filled: #005cc7;
```
djcouto, line 11: `#{kv-color('primary')}`

djcouto (approved):
:shipit:

pedrorebelocosta (approved):

packages/ui-components/src/components/range/range.scss
GoncaloCamaz: @djcouto will update

PR #132, opened 2022-08-17
fix(multi-select-dropdown): emit new object with new selected options

nelsonmestevao:


Insatisfeito (approved):

packages/ui-components/src/components/multi-select-dropdown/multi-select-dropdown.tsx
```diff
@@ -59,10 +59,9 @@ export class KvMultiSelectDropdown implements IMultiSelectDropdown, IMultiSelect
 	private selectOption = (event: CustomEvent<string>) => {
 		const option = event.detail;
 		// triple bang for handling false, true and undefined
-		this.selectedOptions[option] = !!!this.selectedOptions[option];
-		this.calculateLabelValue();
```
nelsonmestevao, line 63: Since we are no longer mutating `this.selectedOptions`, we can only calculate the label value when those are updated by the parent's component.

miguelpinto98 (approved):

djcouto (approved):
:shipit:

PR #134, opened 2022-08-19
fix(action button): remove focus height & width 100%

laraujo7:


GoncaloCamaz (approved):

nelsonmestevao (approved):

djcouto (approved):

miguelpinto98 (changes requested):
This is a bypass solution. We should try to fix this behaviour correctly.

laraujo7:
> This is a bypass solution. We should try to fix this behaviour correctly.

What I think it's happening is that the button doesn't have the props 'height: 100%' and 'width: 100%', but in the focus mode it has, so the button, when it gets focused, is trying to expand. That's why I think the solution is removing this props.

miguelpinto98:
Try to add position absolute to the content, z-index: -1, left: 0. The before content will be the button content, and we see the blur effect working.
Right now, it's only a pixel, that push the button to down.

laraujo7:
I added your suggestion for the focus problem. The only problem was that the left prop couldn't be 0, but it had to be the same as the padding 💪

packages/ui-components/src/components/action-button/action-button.scss
```diff
@@ -232,9 +232,10 @@ $btn-paddings-x: (
 		&::before {
 			content: '';
 			filter: blur(2px);
-			background: $btn-focus-border-color;
-			width: 100%;
-			height: 100%;
+			position: absolute;
+			z-index: -1;
+			inset: 0;
```
miguelpinto98, line 237: 👍

miguelpinto98 (approved):
let's go!

PR #135, opened 2022-08-23
refactor(range): Refactor range & fix range in firefox

laraujo7:
I started this PR because the range wasn't working correctly in firefox. When you tried to move the slider by dragging the button it didn't work.
When trying to fix the problem, I realized that I could do a little refactor and simplify the component.
I tested in firefox & chrome and both were fine, tell me what you think and if I'm doing something wrong or if I can do something better.

### DIFF
--- a/packages/react-ui-components/src/stories/Components/Range.stories.tsx
+++ b/packages/react-ui-components/src/stories/Components/Range.stories.tsx
@@ -1,6 +1,7 @@
 import { ComponentStory } from '@storybook/react';
 import React from 'react';
 import { KvRange } from '../../components';
+import { useArgs } from '@storybook/client-api';
 
 KvRange.displayName = 'KvRange';
 
@@ -22,7 +23,12 @@ export default {
 	}
 };
 
-const RangeTemplate: ComponentStory<typeof KvRange> = args => <KvRange {...args}></KvRange>;
+const RangeTemplate: ComponentStory<typeof KvRange> = args => {
+	const [_, updateArgs] = useArgs();
+	const onValueChange = (event: CustomEvent<number>) => updateArgs({ value: event.detail });
+
+	return <KvRange {...args} onValueChange={onValueChange} />;
+};
 
 export const DefaultState = RangeTemplate.bind(this);
 DefaultState.args = {
--- a/packages/ui-components/src/components/range/range.helper.ts
+++ b/packages/ui-components/src/components/range/range.helper.ts
@@ -1,5 +1,5 @@
-export const getInputPercentageFromValue = (inputValue: string, min: number, max: number) => {
-	return (100 * (parseInt(inputValue) - min)) / (max - min);
+export const getInputPercentageFromValue = (inputValue: number, min: number, max: number) => {
+	return (100 * (inputValue - min)) / (max - min);
 };
 
 export const getOffset = (percentage: number) => {
--- a/packages/ui-components/src/components/range/range.helper.ts
+++ b/packages/ui-components/src/components/range/range.helper.ts
@@ -3,6 +3,8 @@ export const getInputPercentageFromValue = (inputValue: string, min: number, max
 };
 
 export const getOffset = (percentage: number) => {
-	const invertPercentage = 1 - percentage / 100;
-	return invertPercentage * 10 - 4;
+	const offSet = -0.25;
+	const halfDistance = percentage - 50;
+
+	return halfDistance * offSet;
 };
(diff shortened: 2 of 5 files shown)


packages/ui-components/src/components/range/range.helper.ts
```diff
@@ -1,8 +1,9 @@
-export const getInputPercentageFromValue = (inputValue: string, min: number, max: number) => {
-	return (100 * (parseInt(inputValue) - min)) / (max - min);
+export const getInputPercentageFromValue = (inputValue: number, min: number, max: number) => {
+	return (100 * (inputValue - min)) / (max - min);
 };
 
 export const getOffset = (percentage: number) => {
-	const invertPercentage = 1 - percentage / 100;
-	return invertPercentage * 10 - 4;
+	const offSet = -2.5;
```
miguelpinto98, line 6: can this be on a configuration with more meaning?

packages/ui-components/src/components/range/range.helper.ts
```diff
@@ -1,8 +1,9 @@
...
+
+	return (percentage / 10 - 5) * offSet;
```
miguelpinto98, line 8: same for this `10 - 5`

miguelpinto98 (changes requested):
One little thing here

packages/ui-components/src/components/range/range.tsx
```diff
@@ -49,10 +49,9 @@ export class KvRange implements IRange, IRangeEvents {
 			<Host>
 				<div class="range-container">
 					<input id="slider" class="slider" type="range" min={this.min} max={this.max} value={this.value} step={this.step} onInput={this.applyCssStyles} />
-					<div id="selector" class="selector">
-						<div class="select-btn"></div>
-						<span class="select-value">{this.value}</span>
-					</div>
+					<span id="select-value" class="select-value">
```
GoncaloCamaz, line 52: This div was here due to the selector button design defined in Zeplin. I believe that with this approach you removed the style from the selector button and now the button does work because it is the default range selector from each browser. Please confirm and dont forget to follow the design established in zeplin 😃

GoncaloCamaz (changes requested):

packages/ui-components/src/components/range/range.helper.ts
laraujo7: I needed to calculate the distance between the half of the slider and the button, that's why I subtract 5. I made it more clear in my newest commit.

packages/ui-components/src/components/range/range.helper.ts
laraujo7: Well.. for some reason the label above the button was not following the button properly with only the percentage, I suspect it's because the size of the button was changed, but I'm not sure. To be honest, I got this calculation by trial and error, it seems to work with every input I put.

GoncaloCamaz (approved):

nelsonmestevao (approved):

pedrorebelocosta (approved):

packages/ui-components/src/components/range/range.tsx
```diff
@@ -14,7 +14,7 @@ export class KvRange implements IRange, IRangeEvents {
 	/** @inheritdoc */
 	@Prop({ reflect: true }) max: number;
 	/** @inheritdoc */
-	@Prop({ reflect: true }) value?: number = 0;
+	@Prop({ reflect: true, mutable: true }) value?: number = 0;
```
nelsonmestevao, line 17: As previously discussed, I think we should not mutate this prop and only emit in [`onValueChange`](https://github.com/kelvininc/ui-components/pull/135/files#diff-fffe65616287038846028a9f391b7b789686162c5bc92721fdfce4c956cbcf5cR43).

packages/react-ui-components/src/stories/Components/Range.stories.tsx
```diff
@@ -22,7 +23,12 @@ export default {
 	}
 };
 
-const RangeTemplate: ComponentStory<typeof KvRange> = args => <KvRange {...args}></KvRange>;
+const RangeTemplate: ComponentStory<typeof KvRange> = args => {
+	const [_, updateArgs] = useArgs();
```
djcouto, line 27: Nice 👍

djcouto (approved):

miguelpinto98 (changes requested):
Screenshot 2022-08-30 at 15 34 42
Screenshot 2022-08-30 at 15 35 35

I tested this component in a tryout and I'm seeing the bar bigger than expected, over the circle. The same happens in the storybook if a bigger width is setted on kv-range. Can you check this? 
@laraujo7

laraujo7:
Yeap, that was a bug. It correct now @miguelpinto98 💪

miguelpinto98 (approved):
👍

PR #136, opened 2022-08-25
refactor: replace moment with day.js

nelsonmestevao:
The reasons behind this can be found in [these notion notes](https://www.notion.so/kelvininc/Best-Alternatives-to-Moment-JS-78b124a536dd4622a73d12132bdbb034?p=e4bbaf01b8e841b6974e64a0cb97d0f3&pm=s).

### References

- https://medium.com/swlh/best-moment-js-alternatives-5dfa6861a1eb

- https://inventi.studio/en/blog/why-you-shouldnt-use-moment-js

- https://medium.datadriveninvestor.com/https-medium-com-sabesan96-why-you-should-choose-day-js-instead-of-moment-js-9cf7bb274bbd

- https://ch-k.medium.com/replace-moment-js-with-day-js-74e54a545f96


packages/ui-components/package.json
```diff
@@ -43,6 +43,8 @@
 		"clean": "rm -rf node_modules dist loader unit-coverage www docs",
 		"test": "TZ=UTC stencil test --spec --e2e --no-cache",
 		"test:watch": "TZ=UTC stencil test --spec --e2e --watchAll",
+		"test:spec": "TZ=UTC stencil test --spec --no-cache",
```
djcouto, line 46: Nice, but please change the `test` script to use these two scripts.

`"test": "npm run test:spec && npm run test:e2e"`

djcouto (approved):
Outstanding work! 👏

Thank you for such an improvement on my PR 💪

packages/ui-components/src/components/calendar-range-dates-selector/calendar-range-dates-selector.tsx
```diff
@@ -50,21 +50,21 @@ export class KvCalendarRangeDatesSelector implements ICalendarRangeDatesSelector
...
-			this.selectRangeDates.emit({ event, payload: [formatDatetime(clickedDateMoment.startOf('day'))] });
+			this.selectRangeDates.emit({ event, payload: [formatDateTime(clickedDate.startOf('day'))] });
 			return;
 		}
 
 		// check if clicked date is the same as the selected date
-		if (isDateSame(clickedDateMoment, selectedStartDate)) {
+		if (isDateSame(clickedDate, selectedStartDate)) {
 			// check if end date is not selected
 			if (selectedEndDate === undefined) {
```
pedrorebelocosta, line 66: could this comparison be simplified to `!selectedEndDate`? In his currently open PR @laraujo7 is using lodash's `isNil` method to verify falsy values and asked him if it was regarding linting issues, I'm waiting for the answer but I'm asking this because it's mostly to keep consistency in our boolean assertions

packages/ui-components/src/components/range/range.tsx
```diff
@@ -48,7 +48,7 @@ export class KvRange implements IRange, IRangeEvents {
 		return (
 			<Host>
 				<div class="range-container">
-					<input id="slider" class="slider" type="range" min={this.min} max={this.max} value={this.value} step={this.step} onInput={this.applyCssStyles} />
+					<input id="slider" class="slider" type="range" min={this.min} max={this.max} step={this.step} value={this.value} onInput={this.applyCssStyles} />
```
pedrorebelocosta, line 51: why was the order of the args changed? was it to keep min, max and step together?

packages/ui-components/src/utils/date.helper.ts
```diff
@@ -1,90 +1,104 @@
...
+
+// Constructors
+export const fromISO = (date: string): dayjs.Dayjs => newDate(date);
+export const fromDateFields = (day: number, month: number, year: number): dayjs.Dayjs => newDate({ day, month: month - 1, year });
+export const fromDateInput = (date: DateInput): dayjs.Dayjs => newDate(date);
+
 // Months
-export const getMonthName = (month: number): string => `${moment({ month: month - 1 }).format('MMMM')}`;
-export const getNumberOfDaysInMonth = (month: number, year: number): number => moment({ month: month - 1, year }).daysInMonth();
+export const getMonthName = (month: number): string =>
```
pedrorebelocosta, line 33: can this be a one liner?

packages/ui-components/src/utils/date.helper.ts
```diff
@@ -1,90 +1,104 @@
...
+	`${dayjs()
+		.month(month - 1)
+		.format('MMMM')}`;
+export const getNumberOfDaysInMonth = (month: number, year: number): number =>
+	dayjs()
```
pedrorebelocosta, line 38: I see, you kept formatting consistency in the functions.

packages/ui-components/src/utils/types/intl.d.ts
```diff
@@ -0,0 +1,13 @@
+// TODO: Will not be needed in the future (already supported in Chrome, Firefox and Safari, only in Node.js 18)
```
pedrorebelocosta, line 1: 👍

pedrorebelocosta (changes requested):
Just some questions but other than that LGTM 💪 Great job and definitely an improvement to bundle size :)

packages/ui-components/src/components/calendar-range-dates-selector/calendar-range-dates-selector.tsx
nelsonmestevao: Well, I think it can, indeed. I could use `isNil` as well if it is preferred. Maybe @djcouto should just comment if he has anything against, as I am not fully aware of everything that it can imply.

packages/ui-components/src/utils/date.helper.ts
nelsonmestevao: Actually, it was my code formatter, as it has exactly 101 chars (1 after other projects limited :stuck_out_tongue_closed_eyes: ). I just checked, and this package does not have a `prettier:fix` or `format:fix` script. Should I open a PR adding this?

Other thing, the config as a limit of 180 chars, is that not too long? Should I make it one line since it is under that limit?

packages/react-ui-components/src/stories/foundation/Colors.stories.mdx
```diff
@@ -96,17 +96,18 @@ const neutralColors = {
...
 # Colors
-Color is a ** functional ** and ** emotional ** tool. Our color system communicates brand personality, sets a tone, attracts attention, and indicates importance. In other words, it allows the creation of a web interface more usable and helps us consistent experiences across products.
+
+Color is a **functional** and **emotional** tool. Our color system communicates brand personality, sets a tone, attracts attention, and indicates importance. In other words, it allows the creation of a web interface more usable and helps us consistent experiences across products.
 
 ## Main Colors
 
-Every main color can have these * variations *
-*(variations can only be used to complement the base color)*:
+Every main color can have these *variations* *(variations can only be used to
```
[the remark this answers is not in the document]
nelsonmestevao: Yes, it works :+1: 
image

My editor does not the early break :stuck_out_tongue:

packages/ui-components/src/utils/date.helper.ts
pedrorebelocosta: we have a `lint:fix` on this pkg if that's what you're looking for, or am I missing something here?

pedrorebelocosta (approved):

packages/ui-components/src/components/range/range.tsx
nelsonmestevao: I thought it was because eslint was complaining. However, I just tested it and it is not anymore. I guess I will revert it and let PR #137 deal with warnings not related to my changes :+1:

pedrorebelocosta: I mean, from my point of view the change you made makes it a bit more organized, let's see what everyone else thinks :)

packages/ui-components/src/utils/date.helper.ts
nelsonmestevao: I wasn't aware that `lint:ts:fix` was also formatting the code, thank you :+1: I think it is not taking in account the `.prettierrc.json` about `printWidth` though :thinking:

pedrorebelocosta: we have `eslint-plugin-prettier` and `eslint-config-prettier` installed as dependencies in this project and prettier seems to be reporting warnings and errors, however `printWidth` isn't working like you said, could it have something to do with this [bug](https://github.com/eslint/eslint/issues/11408)?

nelsonmestevao: I think https://github.com/prettier/prettier-eslint could be a better approach than relying on `eslint-plugin-prettier` to do all the formatting. The [README.md](https://github.com/prettier/eslint-plugin-prettier#readme) of `eslint-plugin-prettier` says:

> If your desired formatting does not match Prettier’s output, you should use a different tool such as [prettier-eslint](https://github.com/prettier/prettier-eslint) instead.

pedrorebelocosta: That would run prettier formatting before running eslint, to me it looks like a better approach so if there are no other implications behind the change, we can replace it :D what do you think? @djcouto @miguelpinto98

packages/ui-components/src/components/calendar-range-dates-selector/calendar-range-dates-selector.tsx
djcouto: I'm not a fan of overusing the lodash utilities. This is just an undefined check, don't see a reason for the function call to make the exact same check 🤷‍♂️

packages/ui-components/src/components/range/range.tsx
djcouto: LGTM this way 👍

packages/ui-components/src/components/calendar-range-dates-selector/calendar-range-dates-selector.tsx
pedrorebelocosta: all good for me :)

packages/ui-components/src/utils/date.helper.ts
djcouto: This is out of the scope of this PR. Please open an issue so this can be discussed by @kelvininc/frontend team 👍

packages/ui-components/src/components/calendar-range-dates-selector/calendar-range-dates-selector.tsx
nelsonmestevao: But should I do `!selectedEndDate` instead?

djcouto: It doesn't bother me, go ahead 👍

miguelpinto98 (commented):
Good enhanced here!!!

ricardogoncalves89 (commented):
👌

PR #137, opened 2022-08-25
fix: remove lint warnings

laraujo7:
In this pull request, I didn't touch anything from the calendar because I saw an open PR with a lot of changes in the calendar, any method that it's public and should be private.

Also, there are some warnings on the calendar from 'props that shouldn't use .bind()'. I tried to search for a solutions and I found some good practices for going around it: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md


djcouto (approved):
LGTM 🚀

nelsonmestevao (approved):

packages/ui-components/src/components/utils.ts
```diff
@@ -1,10 +1,11 @@
+import { isNil } from 'lodash-es';
 import { DEFAULT_CONFIG } from '../globals/config';
 import { UIComponentsConfig } from '../types';
 
 export const getConfig = (): UIComponentsConfig => {
 	if (typeof window !== 'undefined') {
 		const KvUiComponents = window.KvUiComponents;
-		if (KvUiComponents && KvUiComponents.config) {
+		if (!isNil(KvUiComponents) && !isNil(KvUiComponents.config)) {
```
pedrorebelocosta, line 8: is there any specific reason why we're changing the boolean expr to use `isNil()`?

lodash's `isNil` method compares a given value with the abstract comparison algorithm (a.k.a loose equality) to check for `undefined` or `null` (both falsy values), is there any other falsy value these can assume, or is this just to fix some warning in the linter? If so, can you write it here? other than that LGTM 💪

pedrorebelocosta (approved):

packages/ui-components/src/components/utils.ts
laraujo7: I'm using the `IsNil()` from loadash exactly because it checks both equalities. But yes it's just to fix some warnings in linter 👌

pedrorebelocosta: :shipit:

PR #140, opened 2022-08-30
build: upgrade angular versions and fix missing export

nelsonmestevao:
When using the `@kelvininc/angular-ui-components` in an angular app I was having the following error:

*NG3004*: Unable to import class ToasterController.
> `src/controllers/toaster/toaster-controller.component.ts:22:14`
error *NG3001*: Unsupported private class `ToasterController`. This class is visible to consumers via `ToasterControllerModule` -> `ToasterController`, but is not exported from the top-level library entrypoint.


packages/angular-ui-components/angular.json
```diff
@@ -27,19 +27,15 @@
...
-						"exclude": [
-							"**/node_modules/**"
-						]
+						"lintFilePatterns": ["src/**/*.ts", "src/**/*.html"]
 					}
 				}
 			}
 		}
 	},
-	"defaultProject": "@kelvininc/angular-ui-components"
```
miguelpinto98, line 44: I think this is necessary to generate the dist in the right way

packages/angular-ui-components/package.json
```diff
@@ -25,49 +25,55 @@
...
 		"build:watch": "ng build --watch",
 		"test": "ng test",
 		"lint": "ng lint",
 		"postbuild": "node .scripts/copy-assets.js"
 	},
 	"peerDependencies": {
-		"@angular/common": ">=11.2.14",
-		"@angular/core": ">=11.2.14",
-		"rxjs": ">=6.6.0"
+		"rxjs": ">=7.5.0"
```
miguelpinto98, line 35: To not mess with the configuration of `angular-client-sdk` we should maintain the rxjs, if possible.

packages/angular-ui-components/package.json
```diff
@@ -25,49 +25,55 @@
...
 		"start": "ng serve",
-		"build": "ng build --prod",
+		"build": "ng build --configuration production",
 		"build:watch": "ng build --watch",
 		"test": "ng test",
 		"lint": "ng lint",
 		"postbuild": "node .scripts/copy-assets.js"
 	},
 	"peerDependencies": {
-		"@angular/common": ">=11.2.14",
```
miguelpinto98, line 35: to reduce the bundle size with should be declared.

packages/angular-ui-components/src/controllers/index.ts
```diff
@@ -1,3 +1,6 @@
+// Components
+export { ToasterControllerComponent } from './toaster/toaster-controller.component';
```
miguelpinto98, line 2: Instead of doing this here. Create an index inside of `/toaster` with all the thing that are needed to export. And export it here.

packages/angular-ui-components/tsconfig.lib.prod.json
```diff
@@ -4,8 +4,8 @@
 	"compilerOptions": {
 		"declarationMap": false
 	},
-	"angularCompilerOptions": {
-		"enableIvy": false
+  "angularCompilerOptions": {
```
miguelpinto98, line 7: identation

packages/angular-ui-components/tsconfig.json
```diff
@@ -16,7 +16,7 @@
 		"experimentalDecorators": true,
 		"moduleResolution": "node",
 		"importHelpers": true,
-		"target": "es2015",
+		"target": "es2020",
 		"module": "es2020",
 		"isolatedModules": true,
 		"lib": [
```
miguelpinto98, line 22: probably we can update the lib `es2018` to `es2020`

packages/angular-ui-components/package.json
```diff
@@ -25,49 +25,55 @@
 	"scripts": {
 		"clean": "rm -rf node_modules dist",
 		"start": "ng serve",
-		"build": "ng build --prod",
+		"build": "ng build --configuration production",
 		"build:watch": "ng build --watch",
```
miguelpinto98, line 29: isn't necessary to add `--configuration development` here?

packages/angular-ui-components/package.json
```diff
@@ -25,49 +25,55 @@
...
 	},
 	"peerDependencies": {
-		"@angular/common": ">=11.2.14",
-		"@angular/core": ">=11.2.14",
-		"rxjs": ">=6.6.0"
+		"rxjs": ">=7.5.0"
 	},
 	"dependencies": {
 		"@kelvininc/ui-components": "0.5.3",
 		"tslib": "^2.0.0"
```
miguelpinto98, line 39: increase to `^2.3.0`

packages/angular-ui-components/package.json
```diff
@@ -25,49 +25,55 @@
...
-		"ng-packagr": "^11.0.0",
+		"ng-packagr": "^14.2.0",
 		"protractor": "~7.0.0",
-		"rxjs": "~6.6.0",
+		"rxjs": "~7.5.0",
 		"ts-node": "~8.3.0",
 		"tslint": "~6.1.0",
-		"typescript": "~4.1.5",
-		"zone.js": "~0.11.3"
+		"typescript": "~4.6.4",
```
miguelpinto98, line 75: I think the most recent versions of angular can go to `~4.7.2`

packages/angular-ui-components/package.json
```diff
@@ -25,49 +25,55 @@
...
+		"@typescript-eslint/parser": "5.29.0",
 		"codelyzer": "^6.0.0",
+		"eslint": "^8.18.0",
 		"fs-extra": "10.0.0",
 		"jasmine-core": "~3.6.0",
 		"jasmine-spec-reporter": "~5.0.0",
-		"karma": "~6.1.0",
+		"karma": "~6.4.0",
 		"karma-chrome-launcher": "~3.1.0",
 		"karma-coverage": "~2.0.3",
```
miguelpinto98, line 67:
```
    "karma-coverage": "~2.2.0",
    "karma-jasmine": "~5.1.0",
    "karma-jasmine-html-reporter": "~2.0.0",
```

packages/angular-ui-components/package.json
```diff
@@ -25,49 +25,55 @@
...
+		"@angular/forms": "^14.2.0",
+		"@angular/router": "^14.2.0",
 		"@types/jasmine": "~3.6.0",
 		"@types/node": "^12.11.1",
+		"@typescript-eslint/eslint-plugin": "5.29.0",
+		"@typescript-eslint/parser": "5.29.0",
 		"codelyzer": "^6.0.0",
+		"eslint": "^8.18.0",
 		"fs-extra": "10.0.0",
 		"jasmine-core": "~3.6.0",
```
miguelpinto98, line 63: `"jasmine-core": "~4.3.0",`

packages/angular-ui-components/package.json
```diff
@@ -25,49 +25,55 @@
...
+		"@angular-eslint/template-parser": "14.0.3",
+		"@angular/animations": "^14.2.0",
+		"@angular/cli": "^14.2.1",
+		"@angular/common": "^14.2.0",
+		"@angular/compiler": "^14.2.0",
+		"@angular/compiler-cli": "^14.2.0",
+		"@angular/core": "^14.2.0",
+		"@angular/forms": "^14.2.0",
+		"@angular/router": "^14.2.0",
 		"@types/jasmine": "~3.6.0",
```
miguelpinto98, line 56: `4.0.0`

packages/angular-ui-components/angular.json
```diff
@@ -27,19 +27,15 @@
 					}
```
miguelpinto98, line 27: we can add `"defaultConfiguration": "production"` to build configuration. With this we can remove from package.json the --configuration.

miguelpinto98 (changes requested):
👍  Practically done. Only some questions and suggestions

packages/angular-ui-components/src/controllers/index.ts
nelsonmestevao: Will do :+1:

nelsonmestevao: Should I do `export * from './toaster/toaster-controller.component'` or `export { ToasterControllerComponent } from './toaster/toaster-controller.component';`?

miguelpinto98: Checking the workaround of the angular libraries, they used to export all

Screenshot 2022-08-30 at 15 42 24

packages/angular-ui-components/angular.json
nelsonmestevao: I am getting this linting information that it is a deprecated value:

image

packages/angular-ui-components/package.json
nelsonmestevao: We don't have a development configuration, and I don't think there is a difference in `tsconfig.lib.prod.json`. Or should something be different?

miguelpinto98: Forget. We don't need this. But by default the angular.json will use `tsconfig.json`. Only when configuration flag is set to `production` will gonna use `tsconfig.lib.prod.json`

packages/angular-ui-components/angular.json
miguelpinto98: 🤔 
> defaultProject workspace option has been deprecated (the default is now the current working directory

Let's see whats happen

nelsonmestevao:
Depending on #145 for build to pass.

packages/angular-ui-components/package.json
```diff
@@ -25,51 +25,56 @@
...
+		"build": "ng build",
 		"build:watch": "ng build --watch",
 		"test": "ng test",
 		"lint": "ng lint",
 		"postbuild": "node .scripts/copy-assets.js"
 	},
 	"peerDependencies": {
-		"@angular/common": ">=11.2.14",
-		"@angular/core": ">=11.2.14",
-		"dayjs": "^1.11.5",
```
miguelpinto98, line 37: `"dayjs": "^1.11.5" ` stills a peerDependency

miguelpinto98 (changes requested):
only one thing with the angular package.json. Check please 🙏

ricardogoncalves89 (approved):
LGTM

miguelpinto98 (approved):
🚀

PR #143, opened 2022-08-30
feat(textfield): add left slot to enable currency text

GoncaloCamaz:
Screenshot 2022-08-30 at 18 47 38

### DIFF
--- a/packages/ui-components/src/components/text-field/text-field.base.scss
+++ b/packages/ui-components/src/components/text-field/text-field.base.scss
@@ -24,7 +24,10 @@
 
 	 * @prop --background-color-default: Background color when state is default.
 	 * @prop --background-color-disabled: Background color when state is disabled.
-	 * @prop --right-slot-width: right slot width.
+	 * @prop --right-slot-padding-right: right slot padding.
+	 * @prop --right-slot-width: right slot width
+	 * @prop --left-slot-padding-left: left slot left padding.
+	 * @prop --left-slot-padding-top: left slot top padding.
 	 */
 
 	--input-height-large: 36px;
@@ -37,7 +40,10 @@
 	--text-color-help-text-error: #{kv-color('error')};
 
 	--border-color-error: #{kv-color('error')};
+	--right-slot-padding-right: #{$spacing-9x};
 	--right-slot-width: 36px;
+	--left-slot-padding-left: #{$spacing-6x};
+	--left-slot-padding-top: 2px;
 }
 
 kv-form-label {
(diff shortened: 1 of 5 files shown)


packages/ui-components/src/components/text-field/text-field.base.scss
```diff
@@ -38,6 +38,7 @@
 
 	--border-color-error: #{kv-color('error')};
 	--right-slot-width: 36px;
+	--left-slot-width: #{$spacing-6x};
```
djcouto, line 41: Should this be padding? 
If so, please change the name to `--left-slot-padding`.
Otherwise, it doesn't make much sense to use our spacing system for the width property, so simply change the value 24px

packages/ui-components/src/components/text-field/text-field.base.scss
```diff
@@ -147,6 +148,26 @@ input {
 		height: 100%;
 		user-select: none;
 		padding-left: $spacing-3x;
+		padding-top: 2px;
+		color: var(--text-color-input-default);
+
+		&.invalid {
+			color: var(--border-color-error);
```
djcouto, line 155: Should it be the border-color or the text-color?

djcouto (changes requested):
Two questions and is ready to go 👍

packages/ui-components/src/components/text-field/text-field.base.scss
GoncaloCamaz: Sure! makes sense! will change

packages/ui-components/src/components/text-field/text-field.base.scss
GoncaloCamaz: It should be the text color 😃  will update

packages/ui-components/src/components/text-field/text-field.base.scss
```diff
@@ -38,6 +40,8 @@
 
 	--border-color-error: #{kv-color('error')};
 	--right-slot-width: 36px;
```
djcouto, line 42: Shoul this also be padding? $spacing-9x?

GoncaloCamaz: It should but i was afraid it would introduce breaking changes. But since its a small fix, it might be changed as well

GoncaloCamaz: @djcouto found this prop being used for a padding-right and a width property. I will divide both into proper naming (width and padding) and will look for breaking changes while adding this to kelvin maps

djcouto (approved):
🚀

PR #144, opened 2022-09-01
fix(icon): remove timestamp from copied SVG of symbols [KMAPS-385]

nelsonmestevao:
Even thought the timestamp was added with the best intentions of being able to see the most recent file on top when listing the public directory, that does not work because every time we deploy we get a new file and our users have to reload the file because its name changed.


djcouto (approved):
Yeah, it's kind of sad to see this change being reverted, our intentions were purest 😂

LGMT 👍

pedrorebelocosta (approved):
Sad to see this go, we'll try and figure something out later on :D

PR #145, opened 2022-09-01
test(calendar): fix cases without missing initial date

nelsonmestevao:


packages/ui-components/src/components/calendar/test/calendar.spec.tsx
```diff
@@ -13,7 +13,7 @@ describe('Calendar (unit tests)', () => {
 		beforeEach(async () => {
 			page = await newSpecPage({
 				components: [KvCalendar],
-				html: '<kv-calendar></kv-calendar>'
+				template: () => <kv-calendar initialDate="2022-08-01"></kv-calendar>
```
djcouto, line 16: Can the date be `2022-04-01`? Just to keep consistency with the other tests.

packages/ui-components/src/components/calendar/test/calendar.spec.tsx
```diff
@@ -152,7 +152,7 @@ describe('Calendar (unit tests)', () => {
 			beforeEach(async () => {
 				page = await newSpecPage({
 					components: [KvCalendar],
-					html: '<kv-calendar max-date="2022-04-01"></<kv-calendar>'
+					html: '<kv-calendar initial-date="2022-05-01" max-date="2022-04-01"></kv-calendar>'
```
djcouto, line 155: Also here 👍

djcouto (changes requested):
Nice, thank you for fixing the tests 👍

Just two comments and is good to go 🚢

djcouto (approved):
🚀

packages/ui-components/src/components/calendar/test/calendar.spec.tsx
```diff
@@ -1,19 +1,16 @@
...
 	let component: KvCalendar;
 
 	describe('when uses default props', () => {
-		const now = new Date();
-
 		beforeEach(async () => {
 			page = await newSpecPage({
 				components: [KvCalendar],
-				html: '<kv-calendar></kv-calendar>'
+				template: () => <kv-calendar initialDate="2022-08-01"></kv-calendar>
```
pedrorebelocosta, line 13: You can use JSX here like you did with `<kv-calendar initialDate="2022-04-01"></kv-calendar>`

I think there's no need to use JSX though, you could use the html prop `initial-date` (as stencil automatically converts prop names) because it's a string and not a complex object.

packages/ui-components/src/components/calendar/test/calendar.spec.tsx
```diff
@@ -36,7 +33,7 @@ describe('Calendar (unit tests)', () => {
 			beforeEach(async () => {
 				page = await newSpecPage({
 					components: [KvCalendar],
-					template: () => <kv-calendar initialDate="2022-04-01" selectedDates={['2022-04-05', '2022-04-06', '2022-04-07']}></kv-calendar>
+					template: () => <kv-calendar initialDate="2022-08-01" selectedDates={['2022-08-05', '2022-08-06', '2022-08-07']}></kv-calendar>
```
pedrorebelocosta, line 36: 👍

packages/ui-components/src/components/calendar/test/calendar.spec.tsx
pedrorebelocosta: I just saw the other usages, keep the consistency of JSX usage :)

pedrorebelocosta (approved):

PR #147, opened 2022-09-05
fix(advanced-date-select): return absolute and relative time by timezone

miguelpinto98:


packages/ui-components/src/components/advanced-date-select-dropdown/advanced-date-select-dropdown.tsx
```diff
@@ -38,7 +38,7 @@ export class KvAdvancedDateSelectDropdown implements IAdvancedDateSelectDropdown
 	/** @inheritdoc */
 	@Prop({ reflect: false }) relativeTimeConfig?: ICalendarAdvanceRelativeTimeConfig;
 	/** @inheritdoc */
-	@Prop({ reflect: false }) selectedTimezone?: string;
+	@Prop({ reflect: false }) selectedTimezone?: string = getDefaultTimezone() ?? 'UTC';
```
djcouto, line 41: Move this `UTC` to config as `FALLBACK_TIMEZONE` or directly to the `getDefaultTimezone()`.

packages/ui-components/src/components/advanced-date-select-dropdown/advanced-date-select-dropdown.tsx
```diff
@@ -70,6 +64,8 @@ export class KvAdvancedDateSelectDropdown implements IAdvancedDateSelectDropdown
 
 	componentWillLoad() {
 		this.handleSelectedTimeChange(this.selectedTime);
+
+		this.internalSelectedTimezone = this.selectedTimezone;
```
djcouto, line 68: You should create a handle to assign the `selectedTimezone` to the `internalSelectedTimezone` when the first changes.

packages/ui-components/src/components/calendar-advanced-date-selector/calendar-advanced-date-selector.helper.ts
```diff
@@ -1,30 +1,41 @@
...
-	const defaultTimezone = getDefaultTimezone();
+export const buildTimezoneByOffset = (timezones: string[]): ITimezoneOffset[] => {
+	return timezones
+		.map(name => {
+			const offset = getTimezoneOffset(name);
+			const label = formatTimezoneName(name);
+
+			return { offset, name, label };
+		})
+		.sort((name1, name2) => name1.offset - name2.offset);
```
djcouto, line 22: You can sort by name, right? 
Since the offset is on the suffix, sorting by name will give exactly the same output and also sort alphabetically by the timezone name.

djcouto (approved):
Great work, LGMT 🚀 

(left only a few comments and suggestions)

packages/ui-components/src/components/calendar-advanced-date-selector/calendar-advanced-date-selector.helper.ts
miguelpinto98: it's not possible. The localCompare don't use correctly the negative values.

djcouto: That makes sense, sorry 👍

GoncaloCamaz (approved):

PR #148, opened 2022-09-06
perf(illustrations): change generation of illustrations to build time

laraujo7:
The premise of this PR comes from a need to reduce the size of the SVG files that the user always has to download when first accessing maps, this could be a problem because this file can grow a lot as the illustrations are added.
One approach made by [telekom](https://github.com/telekom/scale)  was to generate each illustration as a component using a script while building the package. This approach will remove the user's waiting time to download the file but will add some build time, which is a good trade-off. 

I wrapped the new illustration components with our` kv-illustration`  so the component's usage would be the same.

Old version `svg-symbols` file size: 365,4 KiB
This version  `svg-symbols` file size: 91,9 KiB

This means that 75% of the file were illustrations.


pedrorebelocosta:
Just a reminder to change the commit message type :D

ricardogoncalves89 (approved):
LTGM just a few suggestions

GoncaloCamaz (approved):

packages/ui-components/scripts/generate-illustrations.js
```diff
@@ -0,0 +1,105 @@
...
+const fg = require('fast-glob');
+const fs = require('fs-extra');
+const SVGO = require('svgo');
+const { camelCase, startCase, upperFirst, groupBy, map, find } = require('lodash');
+const Handlebars = require('handlebars'); // using Handlebars for consistency with components-sketch
+
+const { parse } = require('svg-parser');
+const toHTML = require('hast-util-to-html');
+
+const svgo = new SVGO({
```
pedrorebelocosta, line 11: can we extract the the SVGO config to a `const`?

packages/ui-components/scripts/generate-illustrations.js
```diff
@@ -0,0 +1,105 @@
+const path = require('path');
+const fg = require('fast-glob');
+const fs = require('fs-extra');
+const SVGO = require('svgo');
+const { camelCase, startCase, upperFirst, groupBy, map, find } = require('lodash');
+const Handlebars = require('handlebars'); // using Handlebars for consistency with components-sketch
```
pedrorebelocosta, line 6: you can remove this comment

packages/ui-components/scripts/generate-illustrations.js
```diff
@@ -0,0 +1,105 @@
...
+const INPUT_GLOB = './src/illustrations/**/*.svg';
+const OUTPUT_PATH = './src/components/illustrations';
+const ILLUSTRATION_TEMPLATE_PATH = './scripts/illustration-component.hbs';
+const EXT = '.svg';
+
+main();
+
+async function main() {
+	/* Get SVG data from source files */
+
```
pedrorebelocosta, line 24: remove whitespace

packages/ui-components/scripts/generate-illustrations.js
```diff
@@ -0,0 +1,105 @@
...
+		return {
+			key,
+			tagName,
+			className,
+			illustration: toHTML(getRoot(parse(defaultItem.data))),
+			viewBox: getViewBox(parse(defaultItem.data))
+		};
+	});
+
+	// HAST https://github.com/syntax-tree/hast
```
pedrorebelocosta, line 70: remove comment

packages/ui-components/scripts/generate-illustrations.js
```diff
@@ -0,0 +1,105 @@
...
+	function getViewBox(tree) {
+		return tree.children[0].properties.viewBox || '0 0 24 24';
+	}
```
pedrorebelocosta, line 73: add line

packages/ui-components/scripts/generate-illustrations.js
```diff
@@ -0,0 +1,105 @@
...
+
+		return {
+			type: 'root',
+			children: children
+		};
+	}
+
+	/* Get component templates */
+
+	const illustrationTemplateText = fs.readFileSync(ILLUSTRATION_TEMPLATE_PATH, {
```
pedrorebelocosta, line 86: Can we change this to `rawIllustrationTemplate`? I believe this would make it a bit clearer as to what data type we're dealing with between these 2 lines where you read the file and compile the handlebars template.

packages/ui-components/scripts/generate-illustrations.js
```diff
@@ -0,0 +1,105 @@
...
+		encoding: 'utf-8'
+	});
+	const templateIllustration = Handlebars.compile(illustrationTemplateText);
```
pedrorebelocosta, line 89: suggestion (if you agree with the one above): change to `illustrationTemplate`

pedrorebelocosta (changes requested):
Great job here, just a few suggestions that we can discuss :)

packages/ui-components/scripts/generate-illustrations.js
laraujo7: Agree :muscle:

packages/ui-components/src/components/illustration/readme.md
```diff
@@ -53,16 +53,10 @@ export const SvgIconExample: React.FC = () => (
...
-| `--illustration-color-dark-55`  | 55% darker version of the illustration base color  |
-| `--illustration-color-light-80` | 80% lighter version of the illustration base color |
-| `--illustration-color-light-90` | 90% lighter version of the illustration base color |
-| `--illustration-height`         | Height of illustration                             |
-| `--illustration-success-color`  | Success color used in the illustration             |
-| `--illustration-width`          | Width of illustration                              |
+| Name                    | Description            |
+| ----------------------- | ---------------------- |
+| `--illustration-height` | Height of illustration |
+| `--illustration-width`  | Width of illustration  |
```
ricardogoncalves89, line 59: please confirm why some css custom properties are not here

ricardogoncalves89 (changes requested):
Just check why the build:packages doesn't generate documentation correctly

packages/ui-components/src/components/illustration/readme.md
laraujo7: That's odd, I'm going to check out :+1:

pedrorebelocosta (approved):

ricardogoncalves89 (approved):

PR #152, opened 2022-09-14
fix(range): add offset to the input value

GoncaloCamaz:
Screenshot 2022-09-14 at 12 15 11
Screenshot 2022-09-14 at 12 15 24

This PR changes the value of the offset applied to the number on top of the range slider. 

This is particularly needed when the range is used inside a modal and the width of the slider is almost the same width of the modal. If the number displayed has at least 4 characters, the number will overflow the space available.

Another approach to consider would be adding a flag to enable/disable this offset but i will leave this question open to the review.


djcouto (approved):
🚀

ricardogoncalves89 (approved):

PR #154, opened 2022-09-16
fix(text-field): remove placeholder 0 on type numeric on hover or focus

GoncaloCamaz:
Removes placeholder added by the input mask and fixed some stories that had misused templates


djcouto (approved):
🚀

ricardogoncalves89 (approved):

PR #156, opened 2022-09-19
feat(tooltip): add disable behaviour

GoncaloCamaz:
This PR adds the disable behaviour to the tooltip, preventing the content from it to show when the tooltip is disabled. This is useful in the use-case where we want to show the tooltip when a certain condition is met.


packages/ui-components/src/components/tooltip/tooltip.tsx
```diff
@@ -22,6 +22,10 @@ export class KvTooltip implements ITooltip {
 	@Prop({ reflect: true }) position?: ETooltipPosition;
 	/** @inheritdoc */
 	@Prop({ reflect: false }) options?: Partial<ComputePositionConfig> = DEFAULT_POSITION_CONFIG;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) disabled?: boolean = false;
+	/** @inheritdoc */
+	@Prop({ reflect: false }) contentElement?: HTMLElement = null;
```
ricardogoncalves89, line 28: What is this input used for?

ricardogoncalves89 (changes requested):

packages/ui-components/src/components/tooltip/tooltip.tsx
GoncaloCamaz: ContentElement input is added to enable the parent component to specify the reference of an element since we have a slot being rendered in this component. However, this was added to follow other implementations in the library since this input, beyond being optional, is not needed for the purpose of this PR which is to disable the opening of the tooltip when a certain condition is met by the parent component.

packages/ui-components/src/components/tooltip/tooltip.tsx
```diff
@@ -30,6 +34,56 @@ export class KvTooltip implements ITooltip {
...
+
+	componentDidRender() {
+		const child = this.getContentElement();
+		const tooltip = this.getTooltipElement();
+
+		if (tooltip === null) {
+			return;
+		}
+
+		if (!this.disabled) {
```
djcouto, line 74: Wouldn't it be easier to check if this condition is met on the `showTooltip`? 🤔

djcouto (approved):
🚀

packages/ui-components/src/components/tooltip/tooltip.tsx
```diff
@@ -30,6 +34,51 @@ export class KvTooltip implements ITooltip {
...
+
+	componentDidRender() {
+		const child = this.getContentElement();
+		const tooltip = this.getTooltipElement();
+
+		if (tooltip === null) {
+			return;
+		}
+
+		child.addEventListener('mouseenter', this.showTooltip);
```
djcouto, line 76: Just one last thing, you need to ensure that this event is properly removed and it's not duplicated.
In order to do that you need two things:
- On `disconnectCallback` remove the event listeners.
- Before assigning these listeners please remove the previous ones.

Also, these create private methods to help you:
- `private listenToEvents(child: HTMLElement)`
- `private unlistenToEvents(child: HTMLElement)`

GoncaloCamaz: Thanks! Will do an update 😃

pedrorebelocosta (approved):

ricardogoncalves89 (approved):

PR #160, opened 2022-09-21
fix: remove state manipulations inside dropdown element

nelsonmestevao:
#155 made it mandatory to all components using Dropdown to handle the `onOpenStateChange` event by them selves.


laraujo7 (approved):

djcouto (approved):
:rocket

PR #164, opened 2022-09-27
feat(range) !: adding customization properties for range component

TheSnakeRekt:


pedrorebelocosta (approved):

packages/ui-components/src/components/range/range.base.scss
```diff
@@ -1,18 +1,23 @@
 @import '../../assets/styles/globals';
 
+
 :host {
 	/**
 	 * @prop --range-width: Width of the range slider
+	 * @prop --range-height: Height of the range slider
+	 * @prop --button-selector-width: Width of the range thumb
+	 * @prop --button-selector-height: Height of the range thumb
 	 * @prop --slider-background-filled: color of the slider when its full
```
pedrorebelocosta, line 10: rename this prop to `range` to keep consistency with the others ;)

packages/ui-components/src/components/range/range.base.scss
```diff
@@ -1,18 +1,23 @@
...
 	 * @prop --slider-background-empty: color of the slider when its empty
```
pedrorebelocosta, line 11: same for this one ;)

packages/ui-components/src/components/range/range.base.scss
pedrorebelocosta, line 10: rename this prop's prefix to `range-` to keep consistency with the others ;)

pedrorebelocosta (changes requested):

packages/ui-components/src/components/range/range.types.ts
```diff
@@ -14,4 +14,6 @@ export interface IRange {
 	value?: number;
 	/** (optional) Range value step */
 	step?: number;
+	/** (optional) Show value label */
```
djcouto, line 17: Hide value label? 😄 
Otherwise, change the prop name to `showLabel`

djcouto (changes requested):
Besides what @pedrorebelocosta pointed out, just one more comment.

After that for me, you're good to go 👍

djcouto (approved):

packages/ui-components/src/components/range/range.tsx
```diff
@@ -36,22 +41,22 @@ export class KvRange implements IRange, IRangeEvents {
 	private applyCssStyles = () => {
 		const rangeInputValue = this.getRangeElement();
 
-		const selector = this.el.shadowRoot.getElementById('select-value');
-
 		const inputValue = this.getInputValue(rangeInputValue);
 		const percentage = getInputPercentageFromValue(inputValue, this.min, this.max);
 		const inputOffSet = getInputOffset(percentage);
 		const valueOffset = getValueOffset(percentage, inputValue);
```
GoncaloCamaz, line 47: I believe this valueOffset is only used inside the !this.hideLabel if. If so, maybe would be better to move this calculation inside the if

packages/ui-components/src/components/range/range.base.scss
```diff
@@ -3,16 +3,20 @@
 :host {
 	/**
 	 * @prop --range-width: Width of the range slider
+	 * @prop --range-height: Height of the range slider
+	 * @prop --range-selector-width: Width of the range thumb
```
ricardogoncalves89, line 7: `--range-selector-width` and `--range-selector-height` can be merged into one like `--range-selector-radius`

GoncaloCamaz (approved):

ricardogoncalves89 (changes requested):
See my comment and the commit should have the break reference because you rename some css variables and it can cause some issues in applications that already use theses variables

ricardogoncalves89 (approved):

pedrorebelocosta (approved):

pedrorebelocosta:
Before merging, make sure your commit message is correct as there is a space between the commit type and description:

current:
`feat(range) !: adding customization properties for range component`

when it should be: 
`feat(range)!: adding customization properties for range component`

miguelpinto98:
@TheSnakeRekt Please fix the commit message and rebase.

PR #163, opened 2022-09-27
feat(tree-dropdown): add bold to dropdown when child is selected [KMAPS-426]

nelsonmestevao:
image
image

PR #166, opened 2022-09-29
fix(tree-dropdown): align child icons [KMAPS-426]

nelsonmestevao:
PR #163 broke the children icon alignment when changed the icons from 24px to 20px.


djcouto (approved):
🚀

ricardogoncalves89 (approved):

PR #173, opened 2022-10-03
feat(range)!: adding customization properties

ricardogoncalves89:
for range component

### DIFF
--- a/packages/ui-components/src/components/range/range.base.scss
+++ b/packages/ui-components/src/components/range/range.base.scss
@@ -3,16 +3,18 @@
 :host {
 	/**
 	 * @prop --range-width: Width of the range slider
+	 * @prop --range-height: Height of the range slider
+	 * @prop --range-selector-radius: Radius of the range thumb
 	 * @prop --slider-background-filled: color of the slider when its full
 	 * @prop --slider-background-empty: color of the slider when its empty
+	 * @prop --thumb-border-color: thumb border color
+	 * @prop --range-label-color: range labels color
+	 * @prop --select-label-color: select labels color
 	*/
 
 	--range-width: 100%;
 	--range-height: 12px;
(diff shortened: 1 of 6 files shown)


packages/ui-components/src/components/range/range.base.scss
```diff
@@ -3,16 +3,18 @@
...
+	 * @prop --select-label-color: select labels color
 	*/
 
 	--range-width: 100%;
 	--range-height: 12px;
-	--button-selector-width: 20px;
-	--button-selector-height: 20px;
-	--slider-background-filled: #{kv-color('primary')};
-	--slider-background-empty: #{kv-color('neutral-9')};
+	--range-selector-radius: 20px;
```
pedrorebelocosta, line 17: 👍

packages/ui-components/src/components/range/range.base.scss
```diff
@@ -37,25 +39,25 @@
...
 	}
 
 	&::-moz-range-thumb {
-		border: $spacing solid kv-color('neutral-0');
-		height: var(--button-selector-height);
-		width: var(--button-selector-width);
-		border-radius: 50%;
+		border: $spacing solid var(--thumb-border-color);
 		background: var(--slider-background-filled);
+		height: var(--range-selector-radius);
```
pedrorebelocosta, line 58: I'm wondering if there's any css prop that could help us in these cases where width and height are the same...

Not necessarily a need but a nice to have if we could, since we're defining size using a radius, I've found that using just the height or width and then using the `aspect-ratio` css property with value `1` would work, but it would still add 2 lines instead of minimising it to just one, so I don't think it's worth it.

pedrorebelocosta (approved):
LGTM :D

miguelpinto98 (approved):

PR #174, opened 2022-10-04
fix(tooltip): check if child is null before `removeEventListeners` [KMAPS-532]

nelsonmestevao:
`child` should be inferred as `HTMLElement | null` because `private getContentElement = (): HTMLElement | null`

image


djcouto (commented):
🚀

miguelpinto98 (approved):

PR #182, opened 2022-11-09
fix(dropdown): add `filteredOptions` to select dropdowns

djcouto:
When searching on a dropdown, we now should use the `filteredOptions` in order for the component not to lose the entire dropdown options context.


pedrorebelocosta:
just to open a discussion regarding the theme: if we were to support filters in dropdowns, would it make sense for this prop to be named `filteredOptions` instead? (I'm framing it as a discussion since dropdown filters are not a requirement at the moment)

djcouto:
> just to open a discussion regarding the theme: if we were to support filters in dropdowns, would it make sense for this prop to be named filteredOptions instead? (I'm framing it as a discussion since dropdown filters are not a requirement at the moment)

I'll change it to filteredOptions, it makes more sense 👍

pedrorebelocosta (approved):

nelsonmestevao (approved):
👏

PR #189, opened 2022-11-20
feat(tooltip): display tooltip only when string is ellipsed

laraujo7:


packages/ui-components/src/components/info-label/info-label.tsx
```diff
@@ -105,7 +105,7 @@ export class KvInfoLabel {
 							<div class="description">
 								{this.description && <div class="text">{this.description}</div>}
 								{this.copyValue && (
-									<kv-tooltip text={this.tooltipConfig.label} position={this.tooltipConfig.position}>
+									<kv-tooltip text={this.tooltipConfig.label} position={this.tooltipConfig.position} display={true}>
```
nelsonmestevao, line 108:
```suggestion
									<kv-tooltip text={this.tooltipConfig.label} position={this.tooltipConfig.position} display>
```

Is this valid?

laraujo7: Yap 😅

packages/ui-components/src/components/text-field/test/text-field.e2e.ts
```diff
@@ -1,212 +0,0 @@
-import { E2EPage, newE2EPage, EventSpy, E2EElement } from '@stencil/core/testing';
```
djcouto, line 1: Why did you remove this?

packages/ui-components/src/components/text-field/test/text-field.spec.tsx
```diff
@@ -1,46 +0,0 @@
-import { SpecPage } from '@stencil/core/internal';
-import { KvTextField } from '../text-field';
-import { newSpecPage } from '@stencil/core/testing';
-
-describe('Text Field (unit tests)', () => {
```
djcouto, line 5: And also this?

packages/ui-components/src/components/tooltip/tooltip.tsx
```diff
@@ -28,6 +28,8 @@ export class KvTooltip implements ITooltip {
 	@Prop({ reflect: true }) disabled?: boolean = false;
 	/** @inheritdoc */
 	@Prop({ reflect: false }) contentElement?: HTMLElement = null;
+	/** @inheritdoc */
+	@Prop({ reflect: false }) display?: boolean = false;
```
djcouto, line 32: Despite this being a good way to handle it, I have a different suggestion, which IMO is cleaner. I think you should have a prop called 'behaviour' that would be either 'click', 'hover' or 'truncate'. Then, the tooltip will only be shown if either one of the following conditions is met:
- Behaviour is 'click' and the tooltip content is clicked.
- Behaviour is 'hover' and the tooltip content is with the mouse over.
- Behaviour is 'truncate', the tooltip content width is not enough and is truncated and it is with the mouse over.

What do you think?

djcouto (changes requested):

packages/ui-components/src/components/tooltip/tooltip.tsx
laraujo7: Yeap, I didn't particularly appreciate how I did this. That's an excellent suggestion I like it a lot  💪

packages/ui-components/src/components/tooltip/tooltip.tsx
```diff
@@ -55,12 +57,23 @@ export class KvTooltip implements ITooltip {
...
-		if (this.delay) {
-			this.timeoutDelayId = window.setTimeout(() => {
-				this.showTooltip();
-			}, this.delay);
-		} else {
-			this.showTooltip();
+		const range = new Range();
+		range.selectNodeContents(this.el);
+
+		switch (this.behaviour) {
```
miguelpinto98, line 63: Since all the cases can use the same resource code, I suggest to change this in a more readable way:

```
if (this.behaviour === ETooltipBehaviour.Truncate) {
	const { width: rectWidth } = range.getBoundingClientRect();
	const { width: elWidth } = this.el.getBoundingClientRect();

	if (rectWidth < elWidth) return;
}

if (this.delay) {
	this.timeoutDelayId = window.setTimeout(() => {
		this.showTooltip();
	}, this.delay);
} else {
	this.showTooltip();
}
```

packages/ui-components/src/components/select/readme.md
```diff
@@ -119,3 +119,5 @@ graph TD;
 ```
 
 ----------------------------------------------
+
```
miguelpinto98, line 122: remove changes?

packages/ui-components/src/components/tooltip/tooltip.tsx
miguelpinto98: I have some doubts about click behaviour. What we intent with it? Only open the tooltip if a click is done in the element? Or it's only to have the cursor point in the element?

miguelpinto98 (changes requested):
Awesome improvement, but I left some suggestions. I think we also need to express this changes in our storybook, so please change the tooltip story with this new behaviours. To this PR, also add some description of the problem.

packages/ui-components/src/components/tooltip/tooltip.tsx
ricardogoncalves89: Be careful, we already have a tooltip for click behavior with name kv-toggle-tip. Please speak with @GoncaloCamaz before startiing.

djcouto (changes requested):
LGMT 👍 

Please just add stories with each tooltip behaviour.

laraujo7:
@miguelpinto98 @ricardogoncalves89 @djcouto I change a little bit the strategy, and right now,  I'm searching for the first HTMLElement that has the ellipsis style, and after that, I'm calculating the difference between the Element found and the inner Element. Let me know what you think :muscle:

packages/ui-components/src/components/tooltip/tooltip.types.ts
```diff
@@ -12,4 +12,11 @@ export interface ITooltip {
 	disabled?: boolean;
 	/** (optional) Content element reference */
 	contentElement?: HTMLElement;
+	/** (optional) Display tooltip */
+	behaviour?: ETooltipBehaviour;
+}
+
+export enum ETooltipBehaviour {
+	Hover = 'hover',
```
djcouto, line 20: Sorry, but this seems a bit odd to me. Since we are not adding a "click" behaviour to the list, seems that the 'hover' and 'truncate' are the same behaviour. The difference is that we use 'truncate' the child container will be ellipsed/truncated.

Can you refactor this so the truncate becomes a boolean prop?

Sorry for the trouble.

packages/ui-components/src/components/utils.ts
```diff
@@ -12,3 +12,19 @@ export const getConfig = (): UIComponentsConfig => {
 
 	return DEFAULT_CONFIG;
 };
+
+export const getCollapsedElement = (elements: NodeListOf<ChildNode>): HTMLElement => {
```
djcouto, line 16: This function signature must be:

```suggestion
export const getCollapsedElement = (elements: NodeListOf<ChildNode>): HTMLElement | undefined => {
```

Correct?

djcouto (changes requested):
LGTM! 👍 

However, I want to you try to change a little thing. It should be quick, nevertheless, sorry for your trouble. 🥲

packages/ui-components/src/components/utils.ts
miguelpinto98, line 16: what you think in clear a little bit this function?

my suggestion:
```
export const getCollapsedElement = (elements: NodeListOf<ChildNode>): HTMLElement | undefined => {
	if (elements.length === 0) return;

	return Array.from(elements).find(node => {
		if (getCssStyle(node as HTMLElement, 'text-overflow') === 'ellipsis') {
			return node;
		}

		return getCollapsedElement(node.childNodes);
	}) as HTMLElement;
};
```

miguelpinto98 (changes requested):
Great improvement too! 🚀  Almost done, only small suggestions to base code.
Other thing, put this code in a alpha version to test with a real use case before the merge.

packages/ui-components/src/components/utils.ts
laraujo7: Won't this return the first parent of the Element found?

packages/ui-components/src/components/tooltip/tooltip.tsx
```diff
@@ -55,6 +58,19 @@ export class KvTooltip implements ITooltip {
 	};
 
 	private showTooltipHandler = () => {
+		if (this.truncate) {
+			const collapsedElement = getCollapsedElement(this.el.childNodes);
```
djcouto, line 62: Maybe it would be a nice thing to move this logic to a helper. Something like, `isElementCollpased(element: HTMLElement): boolean`?

packages/ui-components/src/components/tree-item/readme.md
```diff
@@ -135,42 +135,6 @@ export const TreeItemExample: React.FC = () => (
 | `"children"` | The children container. |
 
 
-## CSS Custom Properties
```
djcouto, line 138: Revert this 🙏

packages/ui-components/src/components/tooltip/readme.md
```diff
@@ -75,13 +76,6 @@ export const TagLetterExample: React.FC = () => (
 | `"content"`   | The tooltip content.   |
 
 
-## CSS Custom Properties
-
-| Name                  | Description                                  |
```
djcouto, line 80: Revert this 🙏

packages/ui-components/src/components/utils.ts
```diff
@@ -12,3 +12,19 @@ export const getConfig = (): UIComponentsConfig => {
 
 	return DEFAULT_CONFIG;
 };
+
+export const getCollapsedElement = (elements: NodeListOf<ChildNode>): HTMLElement | undefined => {
```
djcouto, line 16: I know, everyone has a suggestion 😄 

Here is mine:

```ts
export const getCollapsedElement = (elements: NodeListOf<ChildNode>): HTMLElement | undefined => {
	if (elements.length === 0) {
		return;
	}

	const [firstElement, ...siblings] = elements;

	if (isOverflowEllipsis(firstElement as HTMLElement)) {
		return firstElement;
	}

	return getCollapsedElement(siblings);
}

export const isOverflowEllipsis = (element: HTMLElement): boolean =>
	isNodeText(element) && getCssStyle(element, "text-overflow") === "ellipsis";

export const isNodeText = (element: HTMLElement): boolean =>
  	element.nodeName !== "#text";

export const getCssStyle = (element: HTMLElement, prop: string): string =>
  	window.getComputedStyle(element, null).getPropertyValue(prop);

```

djcouto (changes requested):
Final comments, great work 👍

djcouto (approved):
🚀

packages/ui-components/src/components/tooltip/tooltip.types.ts
```diff
@@ -12,4 +12,6 @@ export interface ITooltip {
 	disabled?: boolean;
 	/** (optional) Content element reference */
 	contentElement?: HTMLElement;
+	/** (optional) Display tooltip */
```
ricardogoncalves89, line 15: Add a better description please.

packages/ui-components/src/components/utils.ts
```diff
@@ -12,3 +12,23 @@ export const getConfig = (): UIComponentsConfig => {
...
+	if (isOverflowEllipsis(firstElement as HTMLElement)) {
+		return firstElement;
+	}
+
+	return getCollapsedElement(Array.from(firstElement.childNodes)) ?? getCollapsedElement(siblings);
+};
+
+export const isOverflowEllipsis = (element: HTMLElement): boolean => isNodeText(element) && getCssStyle(element, 'text-overflow') === 'ellipsis';
+
+export const isNodeText = (element: HTMLElement): boolean => element.nodeName !== '#text';
```
ricardogoncalves89, line 32: Node is a Text node if nodeName is different from `#text`??

ricardogoncalves89 (approved):
LGTM. check the suggestions please

packages/ui-components/src/components/utils.ts
laraujo7: Nope, that's wrong. I'm going to change it, nice catch :muscle:

miguelpinto98 (approved):
🚀

PR #190, opened 2022-11-21
fix(textfield): remove placeholder from inputmask configuration

GoncaloCamaz:
This PR removes the placeholder from the inputmask configuration because this was causing the input to display only the first letter of the placeholder. The placeholder should only be defined in the input html tag


miguelpinto98:
So if we have a text-field with an input mask defined, we can't define a placeholder on it?

GoncaloCamaz:
@miguelpinto98 we can, the placeholder is defined natively in the input HTML tag. The problem here was defining the placeholder we received by props with the inputmask plugin. The solution I found was using the placeholder with an empty string in the inputmask configuration. This will define the empty state of the input as '' in the inputmask plugin and let the native input define the placeholder.

ricardogoncalves89 (approved):

miguelpinto98 (approved):

djcouto (approved):

PR #193, opened 2022-12-13
feat(tooltip): add default delay to tooltip

laraujo7:
So, I talked with Iza to understand if this was a maps only requirement or a library requirement and she confirmed me that it was a library requirement.


packages/ui-components/src/components/tooltip/tooltip.tsx
```diff
@@ -18,7 +18,7 @@ import { isElementCollpased } from './tooltip.utils';
 })
 export class KvTooltip implements ITooltip {
 	/** (optional) Delay to show tooltip in milliseconds. */
-	@Prop({ reflect: true }) delay?: number;
+	@Prop({ reflect: true }) delay?: number = 1000;
```
djcouto, line 21: Maybe move this to constant config?

djcouto (approved):
LGTM 👍

ricardogoncalves89 (approved):

miguelpinto98 (approved):
👍

PR #194, opened 2022-12-14
fix(tooltip): add white space normal

feliciofilipe:
I added this style so the white space does not get set by a CSS class in a inner div. What was happening in some cases was that a inner div a style such as `white-space: nowrap;` and the usage of the tooltip component was not producing the desired results (with the tooltip text leaving the card).

### DIFF
--- a/packages/ui-components/src/components/tooltip/readme.md
+++ b/packages/ui-components/src/components/tooltip/readme.md
@@ -2,6 +2,7 @@
 
 <!-- Auto Generated Below -->
 
+
 ## Usage
 
 ### Angular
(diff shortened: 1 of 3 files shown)


djcouto (approved):

laraujo7 (approved):

GoncaloCamaz (approved):

miguelpinto98:
@feliciofilipe can a description of the problem be added here? To understand if it is a feature or a fix?

feliciofilipe:
@miguelpinto98 I added a description above, thanks for the heads up :pray: reach out if you have any questions

PR #196, opened 2022-12-22
docs: bump on introduction page documentation

GoncaloCamaz:
Screenshot 2022-12-22 at 18 26 30
Screenshot 2022-12-22 at 18 26 36

### DIFF
--- a/packages/react-ui-components/src/stories/foundation/Introduction/Introduction.stories.mdx
+++ b/packages/react-ui-components/src/stories/foundation/Introduction/Introduction.stories.mdx
@@ -67,17 +67,16 @@ import '@kelvininc/react-ui-components/assets/fonts/font-proxima-nova.css';
 Now you are ready to use all available *Kelvin UI Components*. For example:
 
 ```tsx
-import { EActionButtonType, KvActionButton } from '@kelvininc/react-ui-components';
+import { EActionButtonType, KvActionButtonText } from '@kelvininc/react-ui-components';
 
 (...)
 
-<KvActionButton
-	type={EActionButtonType.PrimaryButton}
+<KvActionButtonText
+	type={EActionButtonType.Primary}
 	text="My Button"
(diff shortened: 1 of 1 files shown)


packages/react-ui-components/src/stories/foundation/Introduction/Introduction.stories.mdx
```diff
@@ -185,3 +185,55 @@ import { initialize, StyleMode } from '@kelvininc/[react|angular]-ui-components'
...
+				onClickButton={this.onButtonClick}>
+			</KvActionButtonText>
+		</KvTooltip>
+	);
+}
+```
+
+At the first time that the button is hovered, the tooltip will assume a 1 second delay that is defined in our KvTooltip component. However, if you pass the prop to the tooltip and its value is undefined, Stencil will consider the prop as undefined.
+
+If you want your undefined props to assume a default value, please consider defining it like it`s shown in the example bellow:
```
djcouto, line 217: Suggestion

```suggestion
At first render, the `KvTooltip` will assume its prop default value of 1 second. However, when the `ComponentWrapper` re-renders, this prop value will be set to `undefined` overwriting the previous value of 1 second. This happens because the `KvTooltip` doesn't fall back to the default value when a prop is `undefined`, it simply uses that value when the component is mounted.

If you want to work around this behaviour please consider setting your drilled prop to a default value as it's shown in the example below:
```

djcouto (approved):
📖

packages/react-ui-components/src/stories/foundation/Introduction/Introduction.stories.mdx
GoncaloCamaz: Excellent improve! Will update 👍

miguelpinto98 (approved):
👌

ricardogoncalves89 (approved):

PR #198, opened 2023-01-03
fix(kv-icon): remove height offset from icon and add new tag component

GoncaloCamaz:
Screenshot 2023-01-03 at 16 39 35


packages/ui-components/src/components/tag/tag.tsx
```diff
@@ -0,0 +1,31 @@
...
+@Component({
+	tag: 'kv-tag',
+	styleUrl: 'tag.scss',
+	shadow: true
+})
+export class KvTag {
+	/** Tag text */
+	@Prop({ reflect: true }) text: string;
+	/** (optional) Determines if the Tag has content on its left slot */
+	@Prop({ reflect: true }) hasLeftSlot?: boolean = false;
```
ricardogoncalves89, line 12: You don't need this prop you can know that with 
```
private get hasLeftSlot() {
	return !isNil(this.el.querySelector('[slot="left-slot"]'));
}
```
and add a `left-slot` name to the slot

ricardogoncalves89 (changes requested):

packages/ui-components/src/components/tag/tag.tsx
GoncaloCamaz: Thanks! Will update

ricardogoncalves89 (approved):

packages/ui-components/src/components/icon/icon.scss
```diff
@@ -14,6 +14,7 @@
 }
 
 .icon {
+	display: flex;
```
djcouto, line 17: 🙏

packages/ui-components/src/components/tag/tag.scss
```diff
@@ -0,0 +1,31 @@
...
+	border-radius: 4px;
+	padding: $spacing;
+	background-color: var(--tag-background-color);
+
+	.tag-text {
+		@include kv-font-label-small-uppercase-regular;
+
+		color: var(--tag-text-color);
+
+		&.has-left-slot {
```
djcouto, line 27: You don't need this, just add `gap: $spacing`. It will automatically add a gap between all children.

This way you can remove this has left slot logic from the component.

packages/ui-components/src/components/tag/test/tag.e2e.ts
```diff
@@ -0,0 +1,18 @@
...
+describe('Tag (end-to-end)', () => {
+	let page: E2EPage;
+
+	describe('when it renders with text', () => {
+		beforeEach(async () => {
+			page = await newE2EPage();
+			await page.setContent('<kv-tag text="Default Text"></kv-tag>');
+		});
+
+		it('should render a tag with the text Default Text', async () => {
```
djcouto, line 12: Will it make sense to add a test to when the tag has a slot?

djcouto (changes requested):
Almost 💪

laraujo7 (approved):

packages/ui-components/src/components/tag/test/tag.e2e.ts
GoncaloCamaz: Added test bellow

packages/ui-components/src/components/tag/tag.tsx
```diff
@@ -0,0 +1,22 @@
+import { Component, Host, Prop, h } from '@stencil/core';
+
+@Component({
+	tag: 'kv-tag',
+	styleUrl: 'tag.scss',
+	shadow: true
+})
+export class KvTag {
+	/** Tag text */
```
pedrorebelocosta, line 9: I believe we're missing `(required)` for proper documentation generation

pedrorebelocosta (changes requested):
Great job :D I believe you're just missing one thing in the props but correct me if I'm wrong :)

packages/ui-components/src/components/tag/tag.tsx
GoncaloCamaz: Yeah! nice catch! However, here I will add the (optional) instead. Since we have the slot there, this text might not even be necessary.

pedrorebelocosta (approved):

packages/ui-components/src/components/tag/tag.tsx
```diff
@@ -0,0 +1,22 @@
...
+})
+export class KvTag {
+	/** (optional) Tag text */
+	@Prop({ reflect: true }) text?: string;
+
+	render() {
+		return (
+			<Host>
+				<div class="tag-container">
+					<slot name="left-slot"></slot>
```
djcouto, line 16: Will it make sense to add a `right-slot`? Just in case we need one 😅

djcouto (approved):

packages/ui-components/src/components/tag/tag.tsx
```diff
@@ -0,0 +1,22 @@
+import { Component, Host, Prop, h } from '@stencil/core';
+
+@Component({
+	tag: 'kv-tag',
+	styleUrl: 'tag.scss',
+	shadow: true
+})
+export class KvTag {
+	/** (optional) Tag text */
```
ricardogoncalves89, line 9: Should be optional or required??

packages/ui-components/src/components/tag/tag.tsx
```diff
@@ -0,0 +1,22 @@
...
+	@Prop({ reflect: true }) text?: string;
```
ricardogoncalves89, line 10: I think `label` is a better naming here

packages/ui-components/src/components/tag/tag.tsx
ricardogoncalves89: why is it optional?

packages/ui-components/src/components/tag/tag.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+	render() {
+		return (
+			<Host>
+				<div class="tag-container">
+					<slot name="left-slot"></slot>
+					<div class="tag-text">{this.text}</div>
```
ricardogoncalves89, line 17: if text is optional 
```
{ this.text && (<div ... }

```

ricardogoncalves89: After a rebase with dev you already have a `string.helper.ts` inside `utils` with `isValidLabel` that validate if string is empty or only contain empty spaces. You should use here for validation 
```
{ isValidLabel(this.text) && ... }

```

packages/ui-components/src/components/tag/test/tag.e2e.ts
ricardogoncalves89: and if label was optional  you should add another test without label

ricardogoncalves89 (changes requested):

packages/ui-components/src/components/tag/tag.tsx
GoncaloCamaz: Optional because imagine a use case where the user wants to set a different Typography, He can do it with the slot and discard the usage of the text.

packages/ui-components/src/components/tag/tag.tsx
GoncaloCamaz: I agree

ricardogoncalves89 (approved):

PR #204, opened 2023-01-13
chore: add script to get ui-components bundle size

nelsonmestevao:
### Output preview

image

### References

https://stenciljs.com/docs/config#sourcemap
https://github.com/danvk/source-map-explorer/blob/master/README.md#generating-source-maps
https://github.com/ionic-team/stencil/issues/1744

### DIFF
--- a/packages/ui-components/stencil.config.ts
+++ b/packages/ui-components/stencil.config.ts
@@ -13,6 +13,8 @@ const angularValueAccessorBindings: ValueAccessorConfig[] = [
 	}
 ];
 
+const SOURCE_MAP = process.env.SOURCE_MAP ?? false;
+
 export const config: Config = {
 	namespace: 'Peacock-UI',
 	globalScript: 'src/globals/globals.ts',
@@ -51,6 +53,7 @@ export const config: Config = {
 		}
 	],
 	plugins: [sass()],
+	sourceMap: SOURCE_MAP,
 	testing: {
 		moduleNameMapper: {
 			'^lodash-es$': 'lodash'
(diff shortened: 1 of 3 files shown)


packages/ui-components/stencil.config.ts
```diff
@@ -51,6 +51,7 @@ export const config: Config = {
 		}
 	],
 	plugins: [sass()],
+	sourceMap: true,
```
miguelpinto98, line 54: we should only generate sourceMaps for development purposes.

miguelpinto98 (changes requested):

packages/ui-components/stencil.config.ts
nelsonmestevao: Do you think [this](https://github.com/kelvininc/ui-components/pull/204/commits/507a9bda0e77df01388ba32a0f68443e61087fa6) is a good solution?

nelsonmestevao:
> Nice +1
> 
> Could you add a screenshot of the `source-map-explorer` output? Thank you +1

Thanks :pray: Checkout the description now :raised_hands:

GoncaloCamaz (approved):

packages/ui-components/stencil.config.ts
miguelpinto98: Hmmm, is it possible to use the environment variables to generate the source maps? We can have 2 commands as you do but with the same configuration (conditional to the environment).

nelsonmestevao: I tried to find a way to use the `--dev` for generating the source maps but I didn't find a way in the documentation to create conditional configuration target. Do you know how?

miguelpinto98 (approved):
🚀

miguelpinto98:
Seems good, squash all commits to merge with dev @nelsonmestevao

PR #206, opened 2023-01-18
fix(tag-status)!: replace tag status props to expand functionallity

GoncaloCamaz:
No visual changes were done to the component.
However, to use this component, the user is required to specify the icon to be displayed; the color of the icon defined by the state prop. The label is optional


djcouto (approved):
I like the approach 👍 

I'll also think would be valuable if we create an `AssetTagStatus`, `ClusterTagStatus`, etc, that reuses this component and has an icon and label configs mapped by a state. This way we wouldn't need to replicate these configs every time we need to use this component for those particular use cases. 

But since this is just a personal opinion I'll approve this.

ricardogoncalves89:
> I'll also think would be valuable if we create an `AssetTagStatus`, `ClusterTagStatus`, etc, that reuses this component and has an icon and label configs mapped by a state. This way we wouldn't need to replicate these configs every time we need to use this component for those particular use cases.

@djcouto I think it's more application scope and not design system scope

djcouto:
> > I'll also think would be valuable if we create an `AssetTagStatus`, `ClusterTagStatus`, etc, that reuses this component and has an icon and label configs mapped by a state. This way we wouldn't need to replicate these configs every time we need to use this component for those particular use cases.
> 
> @djcouto I think it's more application scope and not design system scope

You're absolutely right, It doesn't belong in our design system 👍

ricardogoncalves89 (approved):
Please add the `BREAKING CHANGE:` inside the commit message with what you are breaking

packages/ui-components/src/components/tag-status/tag-status.tsx
```diff
@@ -1,23 +1,27 @@
...
 	tag: 'kv-tag-status',
 	styleUrl: 'tag-status.scss',
 	shadow: true
 })
 export class KvTagStatus implements ITagStatus {
-	@Prop({ reflect: true }) type: ETagStatusType = ETagStatusType.Unknown;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) state: ETagState;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) icon: EIconName | EOtherIconName;
```
pedrorebelocosta, line 15: should we make this property non-null using the `!` operator?

pedrorebelocosta (approved):

PR #208, opened 2023-01-19
feat(code-editor): add component

pedrorebelocosta:
image
image

### DIFF
--- a/packages/react-ui-components/package.json
+++ b/packages/react-ui-components/package.json
@@ -50,6 +50,7 @@
 	],
 	"dependencies": {
 		"@kelvininc/ui-components": "0.5.3",
+		"@monaco-editor/react": "^4.4.6",
 		"@rjsf/core": "5.0.0-beta.12",
 		"@rjsf/utils": "5.0.0-beta.12",
 		"@rjsf/validator-ajv8": "5.0.0-beta.12",
@@ -85,6 +86,7 @@
 		"jest": "^23.0.0",
 		"jest-dom": "^3.0.2",
 		"lodash": "^4.17.21",
+		"monaco-editor": "^0.34.1",
 		"prettier": "^2.5.1",
 		"react": "^17.0.2",
 		"react-dom": "^17.0.2",
--- a/packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
+++ b/packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
@@ -1,0 +1,48 @@
+import React, { useCallback, useEffect, useMemo, useRef } from 'react';
+import Editor, { OnMount, useMonaco } from '@monaco-editor/react';
+import type { editor } from 'monaco-editor';
+import { DEFAULT_CHANGE_DEBOUNCE_TIME, DEFAULT_EDITOR_LANGUAGE, DEFAULT_EDITOR_THEME, DEFAULT_MONACO_OPTIONS } from './config';
+import { ECodeEditorTheme, ICodeEditorProps, OnEditorChangeCallback } from './types';
+import { DebouncedFunc, debounce } from 'lodash';
+import { KvLoader } from '../stencil-generated';
+
+export const KvCodeEditor = ({
+	code,
+	readOnly = false,
+	debounceTime = DEFAULT_CHANGE_DEBOUNCE_TIME,
+	loadingComponent = <KvLoader />,
+	language = DEFAULT_EDITOR_LANGUAGE,
+	theme = DEFAULT_EDITOR_THEME,
+	customTheme,
+	onChange
+}: ICodeEditorProps) => {
+	const monaco = useMonaco();
+	const editorRef = useRef<editor.IStandaloneCodeEditor>();
+	const readOnlyRef = useRef<boolean>(readOnly);
+
+	const onChangeDebounced: DebouncedFunc<OnEditorChangeCallback> = useMemo(() => debounce(value => onChange?.(value), debounceTime), []);
+	const onEditorMount: OnMount = useMemo(() => editor => (editorRef.current = editor), []);
+
+	const setupEditor = useCallback(() => {
+		if (!monaco || !customTheme) return;
+		monaco.editor.defineTheme(ECodeEditorTheme.Custom, customTheme);
+		monaco.editor.setTheme(ECodeEditorTheme.Custom);
+	}, [monaco, customTheme]);
+
+	const updateOptions = useCallback(() => {
+		if (!readOnlyRef.current || !editorRef.current) return;
+		readOnlyRef.current = readOnly;
+		editorRef.current.updateOptions({ readOnly });
+	}, [editorRef.current, readOnlyRef.current, readOnly]);
+
+	useEffect(() => {
+		setupEditor();
+		updateOptions();
+	}, [setupEditor]);
+
+	return (
+		<Editor language={language} theme={theme} options={DEFAULT_MONACO_OPTIONS} value={code} loading={loadingComponent} onMount={onEditorMount} onChange={onChangeDebounced} />
+	);
+};
+
+export default KvCodeEditor;
--- a/packages/react-ui-components/src/components/CodeEditor/types.ts
+++ b/packages/react-ui-components/src/components/CodeEditor/types.ts
@@ -1,0 +1,30 @@
+import type { editor } from 'monaco-editor';
+import { ReactNode } from 'react';
+
+export type OnEditorChangeCallback = (value: string | undefined) => void;
+export type CustomThemeData = editor.IStandaloneThemeData;
+
+export enum ECodeEditorTheme {
+	Dark = 'vs-dark',
+	Light = 'light',
+	Custom = 'custom'
+}
+
+export interface ICodeEditorProps {
+	/** Use this property to pass a starting value to the editor. */
+	code?: string;
+	/** Use this property to set the editor to a read only mode where the user cannot modify the value. */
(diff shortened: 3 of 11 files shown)


packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -0,0 +1,44 @@
...
+	const monaco = useMonaco();
+	const editorRef = useRef<editor.IStandaloneCodeEditor>();
+	const readOnlyRef = useRef<boolean>(readOnly);
+
+	const onChangeDebounced: DebouncedFunc<OnEditorChangeCallback> = useMemo(() => debounce(value => onChange?.(value), debounceTime), []);
+	const onEditorMount: OnMount = useMemo(() => editor => (editorRef.current = editor), []);
+
+	useEffect(() => {
+		if (!monaco) return;
+		if (customTheme) {
```
GoncaloCamaz, line 28: Would here make sense move this customTheme to the line above to clean up some code?

`if(!monaco || !customTheme) return;`

And then bellow without the if statement:
`
monaco.editor.defineTheme(ECodeEditorTheme.Custom, customTheme);
monaco.editor.setTheme(ECodeEditorTheme.Custom);
`

packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -0,0 +1,44 @@
...
+			monaco.editor.setTheme(ECodeEditorTheme.Custom);
+		}
+	}, [monaco, customTheme]);
+
+	useEffect(() => {
+		if (readOnlyRef?.current !== readOnly) {
+			readOnlyRef.current = readOnly;
+		}
+		editorRef.current?.updateOptions({ readOnly });
+	}, [editorRef.current, readOnly]);
```
GoncaloCamaz, line 39: Does this readonly control method work? Could not check via storybook. Why did you choose to follow this approach instead of using the options prop from the Editor?

GoncaloCamaz (changes requested):
Great work here! 💪  Just one small fix and one question regarding the readOnly implementation.

packages/react-ui-components/package.json
```diff
@@ -50,11 +50,13 @@
...
 	"dependencies": {
 		"@kelvininc/ui-components": "0.5.3",
+		"@monaco-editor/react": "^4.4.6",
 		"@rjsf/core": "5.0.0-beta.12",
 		"@rjsf/utils": "5.0.0-beta.12",
 		"@rjsf/validator-ajv8": "5.0.0-beta.12",
 		"classnames": "^2.3.1",
-		"fs-extra": "10.0.0"
+		"fs-extra": "10.0.0",
+		"monaco-editor": "^0.34.1"
```
djcouto, line 59: isn't this a dev-dependency since we are using this for its types?

packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
pedrorebelocosta: Hmm I'd done this before but the linter would throw an error, I must've missed the not operator and got confused. Thanks for spotting it, I'll go ahead and change it :D

packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
pedrorebelocosta: This is a tricky one, the monaco editor allows you to pass the readOnly prop in it's constructor options, but then to update it you must use the `updateOptions` function, this wrapper gives us a few [options](https://www.npmjs.com/package/@monaco-editor/react#editor-instance) to retrieve the editor's instance after rendering it, and this approach seemed like the most logical to me when it comes to preventing unnecessary re-renders.

Regarding the storybook responsiveness, I'm not sure what you meant, the control appears to work fine, is there something I'm missing here?

Let me know what you think :)

packages/react-ui-components/package.json
pedrorebelocosta: you're right, changing it :D

packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
GoncaloCamaz: Yep sounds good! The read only prop was not working on my side but it seems its ok, maybe I was being 👜  😄

packages/react-ui-components/src/stories/Components/CodeEditor.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+		</div>
+	);
+};
+
+export const Default = DefaultCodeEditorTemplate.bind(this);
+
+const CustomCodeEditorTemplate: ComponentStory<typeof KvCodeEditor> = args => {
+	const wrapperStyle = { width: '640px', height: '480px' };
+
+	const customDarkTheme: editor.IStandaloneThemeData = {
```
ricardogoncalves89, line 66:
```suggestion
	const customDarkTheme: CustomThemeData = {
```

ricardogoncalves89 (changes requested):
Good jog here💪. Just a few suggestions

djcouto (changes requested):
Some comments 💪

packages/react-ui-components/src/components/CodeEditor/types.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export enum ECodeEditorTheme {
+	Dark = 'vs-dark',
+	Light = 'light',
+	Custom = 'custom'
+}
+
+export interface ICodeEditorProps {
+	code?: string;
```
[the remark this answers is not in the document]
pedrorebelocosta: can you check if want you want is on `readme.md`? or do you mean typedoc?

packages/react-ui-components/src/components/CodeEditor/types.ts
```diff
@@ -0,0 +1,27 @@
...
+export type OnEditorChangeCallback = (value: string | undefined) => void;
+export type CustomThemeData = editor.IStandaloneThemeData;
+
+export enum ECodeEditorMode {
+	YAML = 'yaml',
+	JSON = 'json'
+}
+
+export enum ECodeEditorTheme {
+	Dark = 'vs-dark',
```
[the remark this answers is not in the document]
pedrorebelocosta: These are coupled to monaco's default themes, hence the name 'vs-dark' (docs [here](https://microsoft.github.io/monaco-editor/api/modules/monaco.editor.html#BuiltinTheme)), there are only 4 default themes:  `"vs" | "vs-dark" | "hc-black" | "hc-light"`.

packages/react-ui-components/src/components/CodeEditor/types.ts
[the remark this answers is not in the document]
djcouto: I mean here, like a [doc comment](https://typedoc.org/guides/doccomments/#doc-comments) 😄

packages/react-ui-components/src/components/CodeEditor/types.ts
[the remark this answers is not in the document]
djcouto: Oh, I didn't get that you were mapping this directly with monaco's themes. It's ok then 👍

packages/react-ui-components/src/components/CodeEditor/types.ts
```diff
@@ -0,0 +1,27 @@
+import { editor } from 'monaco-editor';
+import { ReactNode } from 'react';
+
+export type OnEditorChangeCallback = (value: string | undefined) => void;
+export type CustomThemeData = editor.IStandaloneThemeData;
+
+export enum ECodeEditorMode {
```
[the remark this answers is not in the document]
pedrorebelocosta: There are no types / enums we can extend from, having checked the [docs](https://microsoft.github.io/monaco-editor/api/modules/monaco.languages.html) and the [repo](https://github.com/microsoft/monaco-editor/tree/main/src/basic-languages) I didn't find any, I'll leave this open for discussion @djcouto.

packages/react-ui-components/src/components/CodeEditor/types.ts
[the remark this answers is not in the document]
pedrorebelocosta: I'll add it :)

packages/react-ui-components/src/components/CodeEditor/types.ts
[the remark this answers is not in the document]
djcouto: I don't think "monaco editor" has an enum or type with all its supported languages. IMO and based on the fact that they accept any string as language, we should forward the component language prop to the editor component as a string without validations. If the editor doesn't support the language passed as a string, it will simply not be highlighted and won't raise an error.

packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -0,0 +1,46 @@
...
+	}, [monaco, customTheme]);
+
+	const updateOptions = useCallback(() => {
+		if (!readOnlyRef.current || !editorRef.current) return;
+		readOnlyRef.current = readOnly;
+		editorRef.current.updateOptions({ readOnly });
+	}, [editorRef.current, readOnlyRef.current, readOnly]);
+
+	useEffect(() => {
+		setupEditor();
```
djcouto, line 39: 👌

djcouto (approved):
🚀

ricardogoncalves89 (approved):
🚀

PR #212, opened 2023-01-23
feat: add view column icons

feliciofilipe:
<div>
  Screenshot 2023-01-27 at 10 10 23
  Screenshot 2023-01-26 at 18 43 31
  Screenshot 2023-01-26 at 18 44 00
  Screenshot 2023-01-26 at 18 43 46
</div>


packages/ui-components/src/components/icon/icon.types.ts
```diff
@@ -148,6 +148,8 @@ export enum EIconName {
 	Upgrade = 'kv-upgrade',
 	Upload = 'kv-upload',
 	Uptime = 'kv-uptime',
+	ViewColumn = 'kv-view-column',
+	ViewColumn2 = 'kv-view-column-2',
```
djcouto, line 152: These names and not very explicit.

djcouto (changes requested):
Can you please add a screenshot of the icons you are adding?

djcouto (approved):

packages/ui-components/src/components/icon/icon.types.ts
```diff
@@ -130,6 +133,8 @@ export enum EIconName {
 	SortZa = 'kv-sort-za',
 	Square = 'kv-square',
 	Status = 'kv-status',
+	StarActive = 'kv-star-active',
+	StarInactive = 'kv-star-inactive',
```
ricardogoncalves89, line 137: the icon is the same with different opacity right? that opacity can be added by css you don't need to have different icons

ricardogoncalves89 (changes requested):

packages/ui-components/src/components/icon/icon.types.ts
djcouto: Yes, but is used as an icon. I mean, in every single place where you want to use an inactive start you must pass this opacity setting. That's why I choose to make it an icon.

ricardogoncalves89: that way we should have an inactive state on all icons. If you really need this feature, I suggest adding a property to `kv-icon` to set the inactive state.

djcouto: I see your point, but we cannot be sure that all inactive icons will all the same state.

In this case, I'll remove the inactive icon svg and style it as need 👍

ricardogoncalves89 (approved):

PR #218, opened 2023-01-30
fix(switch-button): replace label with slots

djcouto:


packages/ui-components/src/components/switch-button/switch-button.scss
```diff
@@ -49,7 +49,7 @@ $icon-sizes: (
 
 		kv-icon {
 			--icon-width: #{$icon-size};
-			--icon-height: unset;
+			--icon-height: #{$icon-size};
```
GoncaloCamaz, line 52: 🙏

GoncaloCamaz (approved):
Looks good to me! Thank you for the unset fix 😃 
Should we mark this PR as breaking change?

packages/ui-components/src/components/switch-button/switch-button.tsx
```diff
@@ -58,25 +45,22 @@ export class KvSwitchButton {
...
+					}}
+					part="button"
+					onClick={this.onSwitchClick}
+				>
+					<div class="icon-square" part="icon-square">
+						<kv-icon name={iconName} part="icon-svg" />
 					</div>
 				</div>
+				<slot name="right-slot" />
 			</Host>
```
miguelpinto98, line 64: Should the right/left slots be suitable for changing the checkbox state?
Normally when we click on the switch label, it changes the check state.

miguelpinto98 (changes requested):
I only have some doubts about the switch behavior

packages/ui-components/src/components/switch-button/test/switch-button.e2e.ts
```diff
@@ -98,7 +80,7 @@ describe('Switch Button (end-to-end)', () => {
 				const buttonElement = await page.find('kv-switch-button >>> .switch-button');
 				await buttonElement.click();
 
-				await page.waitForTimeout(300);
+				await new Promise(resolve => setTimeout(resolve, 300));
```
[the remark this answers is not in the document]
djcouto: The switch button has a throttle of 300ms.We need to wait that amount of time before making assertions. I don't think `waitForChanges ` is a good fit here.

packages/ui-components/src/components/switch-button/switch-button.tsx
djcouto: Yes, it's true, but this will be handled by the parent that manages these slots, not in these components.

PR #220, opened 2023-01-30
feat(wizard-header): add component

pedrorebelocosta:
Default:
image

### DIFF
--- a/packages/ui-components/src/types.ts
+++ b/packages/ui-components/src/types.ts
@@ -36,3 +36,4 @@ export * from './components/tooltip/tooltip.types';
 export * from './components/tag-status/tag-status.types';
 export * from './components/tag-alarm/tag-alarm.types';
 export * from './components/switch-button/switch-button.types';
+export * from './components/wizard-header/wizard-header.types';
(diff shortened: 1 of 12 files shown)


ricardogoncalves89 (approved):

packages/ui-components/src/types.ts
```diff
@@ -35,3 +35,4 @@ export * from './components/toggle-tip/toggle-tip.types';
 export * from './components/tooltip/tooltip.types';
 export * from './components/tag-status/tag-status.types';
 export * from './components/tag-alarm/tag-alarm.types';
+export * from './components/wizard-header/wizard-header.types';
```
djcouto, line 38: Don't forget to add the angular component.

pedrorebelocosta: Adding a wizard here is indeed the plan :P I have to make some changes to this component due to the redesign.
Thanks for pointing out the lack of export on the angular lib, totally missed that.

ricardogoncalves89:
>

ricardogoncalves89:
> Are we planning to add a wizard to the UI components? If not, maybe we should generalize this component name 😄
> 
> Also, left you a comment.

Yes @djcouto, I will create a KvWizard here.

djcouto (approved):

PR #221, opened 2023-01-31
fix(modal)!: click outside behavior can now be customized

pedrorebelocosta:
The modal component suffered from some logic issues that wouldn't allow us to define if we wanted a click outside to send a `close` event, this PR introduces a fix to define if we want that behaviour to occur or not.

This PR also introduces a smaller fix to the modal's z-index property so it can be properly rendered above any content, the value can be customised through a CSS prop.

### DIFF
--- a/packages/ui-components/src/components/modal/modal.tsx
+++ b/packages/ui-components/src/components/modal/modal.tsx
@@ -14,9 +14,9 @@ export class KvModal implements IModalConfig, IModalEvents {
 	/** @inheritdoc */
 	@Prop() headerTitle?: string;
 	/** @inheritdoc */
-	@Prop() hasBackdrop: boolean = true;
+	@Prop() showOverlay: boolean = true;
 	/** @inheritdoc */
-	@Prop() closable?: boolean = true;
+	@Prop() closeOnOverlayClick?: boolean = true;
 	/** @inheritdoc */
 	@Prop() showCloseButton: boolean = true;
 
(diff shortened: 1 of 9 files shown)


ricardogoncalves89 (approved):

packages/ui-components/src/components/modal/modal.tsx
```diff
@@ -16,7 +16,7 @@ export class KvModal implements IModalConfig, IModalEvents {
 	/** @inheritdoc */
 	@Prop() hasBackdrop: boolean = true;
 	/** @inheritdoc */
-	@Prop() closable?: boolean = true;
+	@Prop() closeOnOverlayClick?: boolean = true;
```
djcouto, line 19: Is there a use case where you don't what this to happen? I mean, clicking on the overlay and the modal doesn't close?

Also, I don't this is the prop you want to change. This prop is related to if the component is closable by either clicking on the overlay or by clicking on the X button.

djcouto (changes requested):

packages/ui-components/src/components/modal/modal.tsx
djcouto: If your intention here is to reuse the modal component to create a "wizard" flow and you want to prevent the user from accidentally closing this modal by clicking on the overlay container, then I don't think that the overlay should exist because it will have different behaviours on different modals. But, it's not a problem with your approach I guess.

Just please coherency, do the following:
- `hasBackdrop` should be `showOverlay` in order to be coherent with the CSS class and the `showCloseButton` prop.

pedrorebelocosta: makes sense, changed it :)

GoncaloCamaz (approved):

djcouto (approved):

PR #228, opened 2023-02-08
feat(alert): add component

pedrorebelocosta:
image
image
image
image

The `size` property allows you to shift from a row (size large) to a column (size small) flex layout


laraujo7 (approved):

laraujo7:
Actually, I just stumbled upon this use case. Can you make the buttons optional? 
2023-02-10-155814_620x238_scrot

laraujo7 (changes requested):
Actually, I just stumbled upon this use case. Can you make the buttons optional?
2023-02-10-155814_620x238_scrot

packages/ui-components/src/components/alert/alert.tsx
```diff
@@ -0,0 +1,82 @@
...
+					}}
+				>
+					<div class="information">
+						{this.showIcon && <kv-icon name={ALERT_ICON_NAMES[this.type]} class="icon" />}
+						<div class="text-container">
+							<div class="title">{this.label}</div>
+							{this.description && <div class="description">{this.description}</div>}
+						</div>
+					</div>
+					<div class="buttons">
```
miguelpinto98, line 62: Should this be more customizable? Imagine that for a new use case, we need to have a third button, we gonna add another `kv-action-button-text` here? For this purpose why not receive a list of buttons to draw with an associated configuration & callbacks for button click? In last case, having a slot here.

miguelpinto98 (changes requested):
💪  The code seems well divided & created. My thoughts are more about the future possibilities of this component if we need to have for example one more button.
Another thing that occurs to me, is the description part. Imagine we want to pass a list of things to delete, we can only have a simple string. Off course this can be handled in the future, but it's better to start the discussion now. 😄

packages/ui-components/src/components/alert/alert.tsx
pedrorebelocosta: Thanks for the feedback 🙏 💪 

In regards to the buttons imo both approaches are valid, I might be a bit biased toward having the component with a slot as this would allow for more flexibility and would allow us to remove the button styling override from the lib. The approach with a button array still restricts what can be drawn on the right hand side and might make it too clumsy to apply the "frankenstein" button styling, but I haven't tested that yet to draw any conclusions.

About the description... maybe we could just use slots for both? since the component might need that flexibility in the future, we could change the description and buttons to have slots and create UI specific components on the app side?

I'd like to hear the opinion of others

GoncaloCamaz: I would go with slots as well. We can also think of another possibilities in the future here like having other type of buttons or components for this specific part.

GoncaloCamaz (approved):

GoncaloCamaz (commented):
Would the implementation with slots be more adequate here? I can see this component changing in the future.

packages/ui-components/src/components/alert/alert.tsx
pedrorebelocosta: Consensus seems to be that slots are the way to go, I'll change it, if anyone objects please comment here :)

packages/ui-components/src/components/alert/alert.types.ts
```diff
@@ -0,0 +1,37 @@
...
+	Error = 'error',
+	Success = 'success',
+	Warning = 'warning'
+}
+
+export interface IAlertConfig {
+	/** (required) Defines the type of alert to show, possible values are defined in `EAlertType` */
+	type: EAlertType;
+	/** (optional) Defines the size of the component, defaults to EComponentType.Large */
+	size: EComponentSize;
```
djcouto, line 15: you need to add the question mark after all the required fields otherwise they will all be required.

djcouto (changes requested):

packages/ui-components/src/components/alert/alert.types.ts
pedrorebelocosta: 👀

laraujo7 (approved):

packages/ui-components/src/components/alert/alert.config.ts
```diff
@@ -0,0 +1,9 @@
+import { EIconName } from '../icon/icon.types';
+import { EAlertType } from './alert.types';
+
+export const ALERT_ICON_NAMES: { [key in EAlertType]: EIconName } = {
+	[EAlertType.Info]: EIconName.Error,
```
djcouto, line 5: Is this correct?

djcouto (changes requested):

packages/ui-components/src/components/alert/alert.config.ts
pedrorebelocosta: The icon naming is a bit confusing but yes, this is correct, the icon is the same in the case of an Error or Information alert, only thing that changes is the fill color :P

packages/ui-components/src/components/alert/alert.tsx
```diff
@@ -0,0 +1,45 @@
...
+	@Prop({ reflect: true }) description?: string;
+
+	render() {
+		return (
+			<Host>
+				<div
+					class={{
+						'alert-container': true,
+						[`--type-${this.type}`]: true,
+						'small': this.size === EComponentSize.Small
```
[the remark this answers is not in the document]
pedrorebelocosta: Thanks! I'll change it as soon as I clarify the colors issue with the design team ;)

djcouto (approved):
💪

miguelpinto98 (approved):

PR #232, opened 2023-02-13
fix(modal): export modal overlay types

GoncaloCamaz:
Exports the ModalOverlay and its types from its content folder to avoid importing the modal type like its shown bellow:

import { IModalController } from '@kelvininc/react-ui-components/dist/types/components/ModalOverlay/types';


miguelpinto98 (approved):
👍

djcouto (approved):

pedrorebelocosta (approved):

PR #237, opened 2023-02-16
refactor(action-button-text): add Action button text types

laraujo7:


GoncaloCamaz (approved):

djcouto (changes requested):
Nice improvement 💪 
Could you do this to other buttons? `KvActionButtonSplit` and `KvActionButtonIcon`.

packages/ui-components/src/components/action-button-icon/action-button-icon.tsx
```diff
@@ -1,17 +1,17 @@
...
+import { IActionButtonTextConfig } from './action-button-icon-types';
 
 @Component({
 	tag: 'kv-action-button-icon',
 	styleUrl: 'action-button-icon.scss',
 	shadow: true
 })
-export class KvActionButtonIcon implements IButton, IButtonIcon, IButtonEvents, IAnchor {
+export class KvActionButtonIcon implements IActionButtonTextConfig {
 	/** (required) Button's icon symbol name */
```
djcouto, line 15: You don't need these doc blocks here because they are inherited from the interface.

djcouto (changes requested):

packages/ui-components/src/components/action-button-icon/action-button-icon.tsx
pedrorebelocosta: you might need to use `/** @inheritdoc */`

djcouto (approved):

PR #238, opened 2023-02-17
feat(code-editor): add padding top option

pedrorebelocosta:
This pull request adds the possibility of having a padding at the top of the editor, the screenshot may not be very clear but you can see the blank space at the top.

image

After scrolling down:
image


GoncaloCamaz (approved):

djcouto (approved):
Seems it always seems a bit odd to me when we use JS for styling, but carry on 😄

pedrorebelocosta:
> Seems it always seems a bit odd to me when we use JS for styling, but carry on 😄

I tried to make it CSS only but there were quite a few issues when doing that, mainly with the cursor positioning on the Editor, so I decided to go with this approach.

PR #240, opened 2023-02-21
fix(dropdown): Change dropdown list default width to be the same as its action

laraujo7:
When a select option width is more significant than the dropdown list width, the ellipses are not displayed and instead, the dropdown list expands. This PR makes the dropdown list width the same as its action, by default.

2023-02-21-173856_415x421_scrot

I also had to bump up the floating UI dependency because I saw that `MiddlewareArguments`  was being deprecated to `MiddlewareState`   which I needed to use. [Here are the realeases](https://github.com/floating-ui/floating-ui/releases)

2023-02-21-175706_1142x348_scrot

Also, I wanted to know if this is a good approach because this is going to affect every dropdown or if should I just pass it as an options prop.

**---> This PR also changes the default height for the dropdown list <---**


packages/ui-components/src/components/toggle-tip/test/__snapshots__/toggle-tip.spec.tsx.snap
```diff
@@ -6,7 +6,7 @@ exports[`Toggle Tip (unit tests) when has a position should match the snapshot 1
     <div class="toggle-tip-open-element-wrapper" id="toggle-tip-open-element-wrapper">
       <slot name="open-element-slot"></slot>
     </div>
-    <div class="toggle-tip-container" id="toggle-tip-container" part="toggle-tip-container" style="left: -5px; top: NaNpx;">
+    <div class="toggle-tip-container" id="toggle-tip-container" part="toggle-tip-container">
```
GoncaloCamaz, line 9: 👍

packages/ui-components/src/components/dropdown/dropdown.config.ts
```diff
@@ -7,3 +8,17 @@ export const DEFAULT_INPUT_CONFIG: Partial<ITextField> = {
 	type: EInputFieldType.Text,
 	state: EValidationState.None
 };
+
+export const DEFAULT_DROPDOWN_POSITION_CONFIG: Partial<ComputePositionConfig> = {
+	placement: 'bottom',
+	middleware: [
+		offset(8),
```
GoncaloCamaz, line 15: Should we describe this 8 value in a constant?

GoncaloCamaz (approved):

djcouto (approved):

PR #246, opened 2023-03-01
feat(tooltip): add white space variable to tooltip configurable css p…

GoncaloCamaz:
…roperties


laraujo7 (approved):

packages/ui-components/src/components/tooltip/tooltip.scss
```diff
@@ -3,9 +3,11 @@
 :host {
 	/**
 	  * @prop --container-z-index: The z-index value for the tooltip container.
+	  * @prop --container-white-space: The white space strategy for the tooltip container.
 	 */
 
 	--container-z-index: 10;
+	--container-white-space: 'normal';
```
djcouto, line 10: I think you want to use the normal value here, not the string 'normal'.

djcouto (changes requested):

PR #247, opened 2023-03-01
fix(toggle-button)!: update styling and usage

feliciofilipe:
**Description**

This PR fix a couple things about the `toggle-button` and the `toggle-button-group` components:

- corrects the colors according to the figma design;
- updates the disabled prop on the `toggle-button-group` to a disable config so we can control each button individually instead of the all group.
- standardizes the name of both components to match with the figma design (radio -> toggle) 
- adds a radio design option to the  `toggle-button`

**Demo**
image
image
image
image


djcouto:
Please add a screenshot to the description 🙏

GoncaloCamaz:
Add the breaking change info and the ! to the commit message please 🙏 😃

miguelpinto98:
> updates the disabled prop on the toggle-button-group to a disable config so we can control each button individually instead of the all group.

I have some mixed feelings about removing this option from the group component. We can achieve the same result, but we need to iterate individually by each toggle item. In some cases, we want to have all the toggle buttons disabled, and iterating over all elements seems to be unnecessary if we maintain the previous behavior. 

@feliciofilipe

djcouto (changes requested):
This is a breaking change, please the `!` after the commit scope.

Apart from that, LGMT.

packages/react-ui-components/src/components/SchemaForm/Widgets/CheckboxesWidget/CheckboxesWidget.tsx
```diff
@@ -43,7 +42,7 @@ const CheckboxesWidget = ({ schema, label, id, disabled, options, value, require
 			{(get(uiSchema, ['ui:title']) || schema.title || label) && (
 				<KvFormLabel id={`${id}-title`} label={get(uiSchema, ['ui:title']) || schema.title || label} required={required} />
 			)}
-			<KvRadioButtonGroup buttons={buttons} selectedButtons={selectedButtons} onCheckedChange={onCheckedChange} disabled={disabled} />
+			<KvToggleButtonGroup buttons={buttons} selectedButtons={selectedButtons} onCheckedChange={onCheckedChange} />
```
miguelpinto98, line 45: Since we removed the previous state of disabled from here, we should add support to it here. Add support to `disabledButtons` .

packages/ui-components/src/components/toggle-button/toggle-button.types.ts
```diff
@@ -2,12 +2,12 @@ import { EventEmitter } from '@stencil/core';
...
 	/** (required) The value to be emitted upon click events */
 	value: string;
-	/** (optional) The button's icon. Only valid for radio button icon */
+	/** (optional) The button's icon. Only valid for toggle button icon */
 	icon?: EIconName | EOtherIconName;
-	/** (optional) The button's label. Only valid for radio button text */
+	/** (optional) The button's label. Only valid for toggle button text */
 	label?: string;
 	/** (optional) Sets the button's styling to be disabled and disables click events */
 	disabled?: boolean;
```
miguelpinto98, line 13: seems this prop `disabled` can be removed from here since isn't used to define this state on the component.

miguelpinto98 (changes requested):
Almost done! All good, but is necessary to have the same functionality to schema form, if necessary to disable buttons.

miguelpinto98:
Screenshot 2023-03-06 at 18 55 41

Also, I noticed that my last toggle usage doesn't follow this new design structure. 
@feliciofilipe check with the design team what they really want.

https://www.figma.com/file/qdvgdPbIEVvdYUtmXnMjQo/01---Data-Explorer?node-id=272%3A6495&t=t6JzRN3v0ukNCyq6-1

packages/ui-components/src/components/toggle-button/toggle-button.types.ts
feliciofilipe: When we are using the `toggle button` component by itself, this is the only way to disable it:
image

feliciofilipe:
> Screenshot 2023-03-06 at 18 55 41
> 
> Also, I noticed that my last toggle usage doesn't follow this new design structure. @feliciofilipe check with the design team what they really want.
> 
> https://www.figma.com/file/qdvgdPbIEVvdYUtmXnMjQo/01---Data-Explorer?node-id=272%3A6495&t=t6JzRN3v0ukNCyq6-1

I followed this specs: https://www.figma.com/file/gnN4hUrN4KNzMY1Kggtitk/Design-System-%5BLibrary%5D?node-id=271-2892&t=uwFzRn2N5Hvjzpo0-0 that seem to be consistent. In the implementation that you just reviewed the disabled background color was incorrect and I've just updated.

Was that the difference that was throwing you off or did you notice anything else? :pray: :pray:

packages/react-ui-components/src/components/SchemaForm/Widgets/CheckboxesWidget/utils.ts
```diff
@@ -48,6 +49,21 @@ export const buildSelectedRadioButtons = <T extends CheckboxOption>(
 	}, {});
 };
 
+export const buildDisabledToggleButtons = (
+	buttons: IToggleButton[],
+	disabled: boolean
+): { [key: string]: boolean } => {
+	const disabledButtons:{[key: string]: boolean} = {}
```
miguelpinto98, line 56:
```
const disabledButtons  = buttons.reduce((acc, button) => {
		acc[button.value] = button.disabled ?? false;

		return acc;
}, {});

...
```

packages/react-ui-components/src/components/SchemaForm/Widgets/CheckboxesWidget/utils.ts
```diff
@@ -48,6 +49,21 @@ export const buildSelectedRadioButtons = <T extends CheckboxOption>(
...
+
+	buttons.forEach(button => {
+		disabledButtons[button.value] = button.disabled ?? false;
+	});
+
+	disabledButtons[ALL_BUTTON_VALUE] = disabled || Object.values(disabledButtons).some(value => value);
```
miguelpinto98, line 62: the disabled flag that was passed initially should only be used here?

packages/ui-components/src/components/toggle-button/toggle-button.types.ts
miguelpinto98: yeap seems legit

miguelpinto98 (commented):
In the case of SchemaForm, I still have doubts about the compatibility for disabled states. Check if we have some use case on Maps the disabled checkbox, to understand better how they are defined.

packages/ui-components/src/components/toggle-button-group/toggle-button-group.tsx
```diff
@@ -0,0 +1,43 @@
...
+	styleUrl: 'toggle-button-group.scss',
+	shadow: true
+})
+export class KvToggleButtonGroup implements IToggleButtonGroup, IToggleButtonGroupEvents {
+	/** @inheritdoc */
+	@Prop({ reflect: true }) buttons: IToggleButton[] = [];
+	/** @inheritdoc */
+	@Prop({ reflect: true }) selectedButtons?: Record<string, boolean> = {};
+	/** @inheritdoc */
+	@Prop({ reflect: true }) disabledButtons?: Record<string, boolean> = {};
```
djcouto, line 16: IMO, you should add a `withRadio` and `disabled` prop so we can easily add a radio button to all buttons/disable all buttons without the need to recreate an object. Furthermore, this prop should override the `disabledButtons` and `radioButtons` props values.

packages/ui-components/src/components/toggle-button/toggle-button.tsx
```diff
@@ -65,24 +67,29 @@ export class KvRadioButton implements IRadioButton, IRadioButtonEvents {
...
 					}}
-					part="radio-button"
+					part="toggle-button"
 					onClick={this.onClick}
 					download={this.download}
 					href={this.href}
 					target={this.target}
 				>
+					{this.withRadio && (
+						<div class="toggle-button-icon" part="toggle-icon">
```
djcouto, line 82: We have a `KvRadio`, can we use it here? If the styles do not match, please fix the component.
Also, revisit this class and part names 😄

djcouto (changes requested):

packages/ui-components/src/components/toggle-button/toggle-button.tsx
feliciofilipe: I'll look into it :pray:

packages/ui-components/src/components/toggle-button-group/toggle-button-group.tsx
feliciofilipe: Seems like sensible additions, I'll implement them :bulb:

packages/react-ui-components/src/components/SchemaForm/Widgets/CheckboxesWidget/utils.ts
feliciofilipe: I updated this util so that here we build the buttons and the disabled prop is passed directly to the `kv-toggle-group` component, so if its value is `true` it will override the config

packages/react-ui-components/src/components/SchemaForm/Widgets/CheckboxesWidget/CheckboxesWidget.tsx
```diff
@@ -43,7 +44,7 @@ const CheckboxesWidget = ({ schema, label, id, disabled, options, value, require
 			{(get(uiSchema, ['ui:title']) || schema.title || label) && (
 				<KvFormLabel id={`${id}-title`} label={get(uiSchema, ['ui:title']) || schema.title || label} required={required} />
 			)}
-			<KvRadioButtonGroup buttons={buttons} selectedButtons={selectedButtons} onCheckedChange={onCheckedChange} disabled={disabled} />
```
djcouto, line 46: You don't need the disabledButtons here anymore.

djcouto (approved):
One minor comment. Apart from that LGTM 👍

packages/react-ui-components/src/components/SchemaForm/Widgets/CheckboxesWidget/CheckboxesWidget.tsx
feliciofilipe: The `disabledButtons` give the ability to have one of the buttons disable and rest not. Even though it is not a foreseeable use case on the short term, it makes the overall component more flexible.

What are your thoughts on this?

djcouto: Oh sorry, I thought this was on a storybook story. It makes total sense 😄

packages/ui-components/src/components/toggle-button-group/toggle-button-group.tsx
```diff
@@ -0,0 +1,47 @@
...
+						value={button.value}
+						label={button.label}
+						icon={button.icon}
+						download={button.download}
+						href={button.href}
+						target={button.target}
+						preventDefault={button.preventDefault}
+						disabled={this.disabled || this.disabledButtons[button.value]}
+						checked={this.selectedButtons[button.value]}
+						withRadio={this.withRadio || this.radioButtons[button.value]}
```
miguelpinto98, line 41: as a component definition, make sense to have only one toggle button with a radio?
Seems like a strange behavior, with a request from design team?

packages/ui-components/src/components/toggle-button-group/toggle-button-group.tsx
```diff
@@ -0,0 +1,47 @@
...
+					></kv-toggle-button>
```
miguelpinto98, line 42: self-close tag?

packages/ui-components/src/components/toggle-button/toggle-button.tsx
```diff
@@ -65,24 +67,25 @@ export class KvRadioButton implements IRadioButton, IRadioButtonEvents {
...
+						'toggle-button--only-icon': hasIcon && !hasLabel
 					}}
-					part="radio-button"
+					part="toggle-button"
 					onClick={this.onClick}
 					download={this.download}
 					href={this.href}
 					target={this.target}
 				>
+					{this.withRadio && <kv-radio checked={this.checked} />}
```
miguelpinto98, line 81: This new definition increases the toggle height when enabled:

Screenshot 2023-03-20 at 22 06 20

packages/ui-components/src/assets/svg-symbols.svg
```diff
@@ -110,16 +110,14 @@
...
 		xmlns="http://www.w3.org/2000/svg">
-		<path d="M0 0h16v16H0V0z" fill="none"/>
-		<path d="M8 0c4.409 0 8 3.582 8 8 0 4.409-3.582 8-8 8-4.409 0-8-3.582-8-8 0-4.409 3.582-8 8-8zm0 1C4.137 1 1 4.14 1 8c0 3.863 3.14 7 7 7 3.863 0 7-3.14 7-7 0-3.863-3.14-7-7-7z"/>
+		<path d="M10 20C8.61667 20 7.31667 19.7373 6.1 19.212C4.88333 18.6873 3.825 17.975 2.925 17.075C2.025 16.175 1.31267 15.1167 0.788 13.9C0.262667 12.6833 0 11.3833 0 10C0 8.61667 0.262667 7.31667 0.788 6.1C1.31267 4.88333 2.025 3.825 2.925 2.925C3.825 2.025 4.88333 1.31233 6.1 0.787C7.31667 0.262333 8.61667 0 10 0C11.3833 0 12.6833 0.262333 13.9 0.787C15.1167 1.31233 16.175 2.025 17.075 2.925C17.975 3.825 18.6873 4.88333 19.212 6.1C19.7373 7.31667 20 8.61667 20 10C20 11.3833 19.7373 12.6833 19.212 13.9C18.6873 15.1167 17.975 16.175 17.075 17.075C16.175 17.975 15.1167 18.6873 13.9 19.212C12.6833 19.7373 11.3833 20 10 20ZM10 18C12.2333 18 14.125 17.225 15.675 15.675C17.225 14.125 18 12.2333 18 10C18 7.76667 17.225 5.875 15.675 4.325C14.125 2.775 12.2333 2 10 2C7.76667 2 5.875 2.775 4.325 4.325C2.775 5.875 2 7.76667 2 10C2 12.2333 2.775 14.125 4.325 15.675C5.875 17.225 7.76667 18 10 18Z" fill="white"/>
```
miguelpinto98, line 115: theses changes are parsed by svgo?

packages/ui-components/src/assets/svg-symbols.svg
```diff
@@ -110,16 +110,14 @@
...
 	</symbol>
 
-	<symbol id="icon-radio-btn-selected" viewBox="0 0 12 12"
+	<symbol id="icon-radio-btn-selected" viewBox="0 0 20 20"
```
miguelpinto98, line 118: also, icon-radio-btn-selected and icon-radio-btn are defined in EOtherIconName, should be moved to EIconName?

miguelpinto98 (changes requested):
Almost done! Only left some questions to resolve before the merge.

packages/ui-components/src/components/toggle-button-group/toggle-button-group.tsx
feliciofilipe: I agree that it does not seem that will have a use case with some buttons with toggle and others without.
On the other hand, this approach offers full flexibility to the developer using the component. What do you think?

packages/ui-components/src/components/toggle-button/toggle-button.tsx
feliciofilipe: Now it should be fixed :pray:

packages/ui-components/src/components/toggle-button-group/toggle-button-group.tsx
miguelpinto98: Ok. I understand the flexibility and we can continue with this approach.
In the future we can think if we are using this and take a better decision.

packages/ui-components/src/components/toggle-button-group/toggle-button-group.tsx
```diff
@@ -0,0 +1,47 @@
...
+					<kv-toggle-button
+						exportparts="toggle-button"
+						value={button.value}
+						label={button.label}
+						icon={button.icon}
+						download={button.download}
+						href={button.href}
+						target={button.target}
+						preventDefault={button.preventDefault}
+						disabled={this.disabled || this.disabledButtons[button.value] || button.disabled}
```
miguelpinto98, line 39: 👍  I think now we have all the possible options

packages/ui-components/src/components/radio/radio.base.scss
```diff
@@ -59,8 +59,9 @@
 	}
 
 	.checked-icon {
-		height: 0;
-		width: 0;
+		height: 8px;
```
miguelpinto98, line 62: you checked this change in other behaviors?

packages/ui-components/src/assets/svg-symbols.svg
```diff
@@ -1233,11 +1221,20 @@
...
 	</symbol>
 
+	<symbol id="kv-radio-btn" viewBox="0 0 24 24"
+		xmlns="http://www.w3.org/2000/svg">
+		<path d="m12 22c-1.3833 0-2.6833-0.2627-3.9-0.788-1.2167-0.5247-2.275-1.237-3.175-2.137s-1.6123-1.9583-2.137-3.175c-0.52533-1.2167-0.788-2.5167-0.788-3.9s0.26267-2.6833 0.788-3.9c0.52467-1.2167 1.237-2.275 2.137-3.175s1.9583-1.6127 3.175-2.138c1.2167-0.52467 2.5167-0.787 3.9-0.787s2.6833 0.26233 3.9 0.787c1.2167 0.52533 2.275 1.238 3.175 2.138s1.6123 1.9583 2.137 3.175c0.5253 1.2167 0.788 2.5167 0.788 3.9s-0.2627 2.6833-0.788 3.9c-0.5247 1.2167-1.237 2.275-2.137 3.175s-1.9583 1.6123-3.175 2.137c-1.2167 0.5253-2.5167 0.788-3.9 0.788zm0-2c2.2333 0 4.125-0.775 5.675-2.325s2.325-3.4417 2.325-5.675c0-2.2333-0.775-4.125-2.325-5.675s-3.4417-2.325-5.675-2.325c-2.2333 0-4.125 0.775-5.675 2.325s-2.325 3.4417-2.325 5.675c0 2.2333 0.775 4.125 2.325 5.675s3.4417 2.325 5.675 2.325z" />
```
miguelpinto98, line 1232: This was parsed by svgo?

packages/ui-components/src/assets/svg-symbols.svg
```diff
@@ -1233,11 +1221,20 @@
...
+	</symbol>
+
+	<symbol id="kv-radio-btn-selected" viewBox="0 0 24 24"
+		xmlns="http://www.w3.org/2000/svg">
+		<path d="m12 17c1.3833 0 2.5627-0.4877 3.538-1.463 0.9747-0.9747 1.462-2.1537 1.462-3.537s-0.4873-2.5627-1.462-3.538c-0.9753-0.97467-2.1547-1.462-3.538-1.462s-2.5623 0.48733-3.537 1.462c-0.97533 0.97533-1.463 2.1547-1.463 3.538s0.48767 2.5623 1.463 3.537c0.97467 0.9753 2.1537 1.463 3.537 1.463zm0 5c-1.3833 0-2.6833-0.2627-3.9-0.788-1.2167-0.5247-2.275-1.237-3.175-2.137s-1.6123-1.9583-2.137-3.175c-0.52533-1.2167-0.788-2.5167-0.788-3.9s0.26267-2.6833 0.788-3.9c0.52467-1.2167 1.237-2.275 2.137-3.175s1.9583-1.6127 3.175-2.138c1.2167-0.52467 2.5167-0.787 3.9-0.787s2.6833 0.26233 3.9 0.787c1.2167 0.52533 2.275 1.238 3.175 2.138s1.6123 1.9583 2.137 3.175c0.5253 1.2167 0.788 2.5167 0.788 3.9s-0.2627 2.6833-0.788 3.9c-0.5247 1.2167-1.237 2.275-2.137 3.175s-1.9583 1.6123-3.175 2.137c-1.2167 0.5253-2.5167 0.788-3.9 0.788zm0-2c2.2333 0 4.125-0.775 5.675-2.325s2.325-3.4417 2.325-5.675c0-2.2333-0.775-4.125-2.325-5.675s-3.4417-2.325-5.675-2.325c-2.2333 0-4.125 0.775-5.675 2.325s-2.325 3.4417-2.325 5.675c0 2.2333 0.775 4.125 2.325 5.675s3.4417 2.325 5.675 2.325z" />
```
miguelpinto98, line 1237: same here?

miguelpinto98 (changes requested):
Please check my comments below

packages/ui-components/src/assets/svg-symbols.svg
feliciofilipe: I used a svg minifier, still came out big like that tho :/

packages/ui-components/src/components/toggle-button/toggle-button.tsx
```diff
@@ -65,24 +67,26 @@ export class KvRadioButton implements IRadioButton, IRadioButtonEvents {
...
+					{/* {this.withRadio && <kv-icon name={this.checked ? EIconName.RadioBtnSelected : EIconName.RadioBtn} />} */}
```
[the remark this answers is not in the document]
feliciofilipe: Nice catch :pray:

packages/ui-components/src/assets/svg-symbols.svg
miguelpinto98: use [svgo](https://github.com/svg/svgo)

feliciofilipe: Now it should be :ok_hand:

packages/ui-components/src/components/radio/radio.base.scss
feliciofilipe: I tested every usage of this in the component library and it seems to hold up:
https://youtu.be/kv09duiA50M

miguelpinto98 (approved):
🚀

PR #251, opened 2023-03-03
feat(select-option)!: add right slot to component

GoncaloCamaz:
BREAKING CHANGE

This PR adds the a slot to the right side of the select option. This option is needed to the new time range picker that displays text on the right of the main label.

Screenshot 2023-03-03 at 16 46 38
Screenshot 2023-03-03 at 16 46 48

### DIFF
--- a/packages/ui-components/src/components/select-option/select-option.scss
+++ b/packages/ui-components/src/components/select-option/select-option.scss
@@ -15,6 +15,7 @@
 	--select-option-height: 32px;
 	--select-option-transition-duration: 100ms;
 	--select-option-flex-alignment: center;
+	--select-option-justify-content: flex-start;
 
 	--select-option-background-color: #{kv-color('neutral-7')};
 	--select-option-background-color-selected: #{kv-color('neutral-6')};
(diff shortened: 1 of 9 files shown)


packages/ui-components/src/components/select-option/select-option.scss
```diff
@@ -28,12 +29,17 @@
 	height: var(--select-option-height);
 	padding: 0 $spacing-4x;
 	display: flex;
-	align-items: var(--select-option-flex-alignment);
```
djcouto, line 31: I'm unsure if these CSS properties should be passed as CSS variables. Instead, they could be added by styling a components part. But this is just my opinion.

packages/ui-components/src/components/select-option/select-option.tsx
```diff
@@ -1,12 +1,15 @@
...
 @Component({
 	tag: 'kv-select-option',
 	styleUrl: 'select-option.scss',
 	shadow: true
 })
 export class KvSelectOption implements ISelectOption, ISelectOptionEvents {
+	@Element() el!: HTMLKvSelectOptionElement;
+
 	/** @inheritdoc */
 	@Prop({ reflect: true }) label!: string;
```
djcouto, line 14: Should we have a prop with the "bottom" and "right" elements?
Otherwise, how would we add a "bottom" or "right" element when using, for example, the `kv-single-select-dropdown` component?

djcouto (changes requested):

PR #256, opened 2023-03-10
feat(relative-time-picker): new relative time picker component

GoncaloCamaz:
Screenshot 2023-03-10 at 00 02 57

### DIFF
--- a/packages/ui-components/src/components/input-wrapper/input-wrapper.scss
+++ b/packages/ui-components/src/components/input-wrapper/input-wrapper.scss
@@ -1,0 +1,44 @@
+@import '../../assets/styles/globals';
+
+:host {
+	display: block;
+}
+
+$input-height: 32px;
+
+.input-container {
+	background-color: kv-color('neutral-7');
+	padding: 0 $spacing-4x;
+	height: $input-height;
+	display: flex;
+	align-items: center;
+
+	kv-icon {
+		visibility: hidden;
+	}
+
+	&--content-hidden {
+		&:hover {
+			background-color: kv-color('neutral-6');
+			cursor: pointer;
+
+			kv-icon {
+				visibility: visible;
+			}
+		}
+	}
+
+	.label-container {
+		@include kv-font-label-small-regular;
+		display: flex;
+		align-items: center;
+		justify-content: space-between;
+		width: 100%;
+
+		color: kv-color('neutral-4');
+	}
+
+	.slot-container {
+		width: 100%;
+	}
+}
--- a/packages/ui-components/src/components/relative-time-picker/relative-time-picker.config.ts
+++ b/packages/ui-components/src/components/relative-time-picker/relative-time-picker.config.ts
@@ -1,0 +1,434 @@
+import { ERelativeTimeComparisonConfig, IRelativeTimePickerOption } from './relative-time-picker.types';
+
+export const TIME_RANGE_UPDATE_INTERVAL = 10000;
+export const CUSTOMIZE_INTERVAL_KEY = 'customize-interval';
+export const CUSTOMIZE_INTERVAL_LABEL = 'Customize Interval';
+
+export const TIMEZONES_PLACEHOLDER = 'Select a timezone';
+export const TIMEZONES_SEARCH_PLACEHOLDER = 'Search for a timezone';
+
+export const DEFAULT_RELATIVE_TIME_OPTIONS_GROUPS: IRelativeTimePickerOption[][] = [
+	[
+		{
+			label: 'Today',
+			value: 'today',
+			comparisonConfig: ERelativeTimeComparisonConfig.RelativeAmountOfUnits,
+			startDate: {
+				amount: 0,
+				unit: 'days'
+			},
+			labelRangeFormatter: {
+				startDateFormatter: 'D MMM',
+				separator: 'until',
+				endDateFormatter: 'HH:mm'
+			}
+		},
+		{
+			label: 'Yesterday',
+			value: 'yesterday',
+			comparisonConfig: ERelativeTimeComparisonConfig.AbsoluteAmountOfUnits,
+			startDate: {
+				amount: -1,
+				unit: 'days'
+			},
+			endDate: {
+				amount: -1,
+				unit: 'days'
+			},
+			labelRangeFormatter: {
+				startDateFormatter: 'D MMM'
+			}
+		},
+		{
+			label: 'This week',
+			value: 'this-week',
+			comparisonConfig: ERelativeTimeComparisonConfig.AbsoluteAmountOfUnits,
+			startDate: {
+				amount: 0,
+				unit: 'week'
+			},
+			endDate: {
+				amount: 0,
+				unit: 'week'
+			},
+			labelRangeFormatter: {
+				startDateFormatter: 'D MMM',
+				separator: 'to',
+				endDateFormatter: 'D MMM'
+			}
+		},
+		{
+			label: 'This month',
+			value: 'this-month',
+			comparisonConfig: ERelativeTimeComparisonConfig.AbsoluteAmountOfUnits,
+			startDate: {
+				amount: 0,
+				unit: 'month'
+			},
+			endDate: {
+				amount: 0,
+				unit: 'month'
+			},
+			labelRangeFormatter: {
+				startDateFormatter: 'D MMM',
+				separator: 'to',
+				endDateFormatter: 'D MMM'
+			}
+		},
+		{
+			label: 'This quarter',
+			value: 'this-quarter',
+			comparisonConfig: ERelativeTimeComparisonConfig.AbsoluteAmountOfUnits,
+			startDate: {
+				amount: 0,
+				unit: 'quarter'
+			},
+			endDate: {
+				amount: 0,
+				unit: 'quarter'
+			},
+			labelRangeFormatter: {
+				startDateFormatter: 'D MMM',
+				separator: 'to',
+				endDateFormatter: 'D MMM YYYY'
+			}
+		},
+		{
+			label: 'This year',
+			value: 'this-year',
+			comparisonConfig: ERelativeTimeComparisonConfig.AbsoluteAmountOfUnits,
+			startDate: {
+				amount: 0,
+				unit: 'year'
+			},
+			endDate: {
+				amount: 0,
+				unit: 'year'
+			},
+			labelRangeFormatter: {
+				startDateFormatter: 'D MMM',
+				separator: 'to',
+				endDateFormatter: 'D MMM YYYY'
+			}
+		}
+	],
+	[
+		{
+			label: 'Last 5 mins',
+			value: 'last-5-m',
+			comparisonConfig: ERelativeTimeComparisonConfig.RelativeAmountOfUnits,
+			startDate: {
+				amount: -5,
+				unit: 'minutes'
+			},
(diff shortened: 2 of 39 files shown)


packages/ui-components/src/components/relative-time-picker/relative-time-picker.scss
```diff
@@ -0,0 +1,41 @@
...
+				--select-option-padding: #{0 $spacing 0 $spacing-4x};
+			}
+		}
+	}
+
+	.selectable {
+		border-top: 1px solid kv-color('neutral-6');
+		padding: $spacing 0;
+
+		.mock {
```
pedrorebelocosta, line 33: is this necessary?

packages/ui-components/src/components/relative-time-picker/relative-time-picker.types.ts
```diff
@@ -0,0 +1,102 @@
...
+	amount?: number;
+	// the unit type to add/subtract to the relative time
+	unit?: string;
+};
+
+export interface IDateTimeRangeFormatter {
+	/** Defines how the initial date is formatted on the relative time select description
+	 * 	If empty, now start date will be displayed
+	 */
+	startDateFormatter?: string;
```
pedrorebelocosta, line 58: do we have any specific way to format these strings? or is this not the purpose of this?

packages/ui-components/src/components/relative-time-picker/test/relative-time-picker.mock.ts
```diff
@@ -0,0 +1,25 @@
+import { ERelativeTimeComparisonConfig, IRelativeTimePickerOption } from '../relative-time-picker.types';
+
+/** Added mock options with fixed dates to prevent tests from breaking due to snapshots differences caused
```
pedrorebelocosta, line 3: 👍

pedrorebelocosta (changes requested):
Massive work on this 💪 small questions and suggestions.

packages/ui-components/src/components/relative-time-picker/relative-time-picker.types.ts
GoncaloCamaz: Yeah, dayJS package provides a method to format a date according with a string. Here we have three formatting strings because in some cases, the start date format is different from the end date format, for example: 13 Feb to 14 Feb 20:30. Since the separator (to, until, ...) can change from option to option just like the start and end date format, I opted by having three different formatters that are concatenated to form the description shown above (ex: 10 Mar until 00:02). 
The formatter accepts determinate date parsing chars like MMM -> Jan, Feb, Mar or MM -> 01, 02, 03. We can have multiple options here.

packages/ui-components/src/components/relative-time-picker/relative-time-picker.scss
GoncaloCamaz: Have to check! Probably not and forgot to remove 👍

packages/ui-components/src/components/relative-time-picker/relative-time-picker.types.ts
pedrorebelocosta: Thanks for the explanation! :)

packages/ui-components/src/components/relative-time-picker/relative-time-picker.config.ts
```diff
@@ -0,0 +1,429 @@
+import { ERelativeTimeComparisonConfig, IRelativeTimePickerOption } from './relative-time-picker.types';
+
+export const TIME_RANGE_UPDATE_INTERVAL = 10000;
+export const CUSTOMIZE_INTERVAL_KEY = 'customize-interval';
+export const CUSTOMIZE_INTERVAL_LABEL = 'Customize Interval';
+export const DROPDOWN_MAX_HEIGHT = '290px';
```
djcouto, line 6: CSS on the JS? 😬

packages/ui-components/src/components/relative-time-picker/relative-time-picker.config.ts
```diff
@@ -0,0 +1,429 @@
...
+
+export const TIMEZONES_PLACEHOLDER = 'Select a timezone';
+export const TIMEZONES_SEARCH_PLACEHOLDER = 'Search for a timezone';
+
+export const DEFAULT_RELATIVE_TIME_OPTIONS_GROUPS: IRelativeTimePickerOption[][] = [
+	[
+		{
```
djcouto, line 13: If we want to select a subset of these options, will we need to define the entire groups? 
Wouldn't it be better if this was indexed by some key and we could select which options we want by passing a set of keys?

packages/ui-components/src/components/relative-time-picker/relative-time-picker.config.ts
```diff
@@ -0,0 +1,429 @@
...
+			},
+			labelRangeFormatter: {
+				startDateFormatter: 'D MMM',
+				separator: 'to',
+				endDateFormatter: 'D MMM'
+			}
+		},
+		{
+			label: 'Last 7 days',
+			value: 'last-7-d',
```
djcouto, line 265: These "values" look like keys to me, probably will be better if we move them to an enum.

packages/ui-components/src/utils/date.helper.ts
```diff
@@ -52,6 +54,7 @@ export const getFirstWeekdayIndexOfMonth = (month: number, year: number): number
 
 // Dates
 export const isDateBefore = (dateA: DateInput, dateB: DateInput): boolean => newDate(dateA).isBefore(dateB, 'days');
+export const isDateTimeBefore = (dateA: DateInput, dateB: DateInput): boolean => newDate(dateA).isBefore(dateB);
```
djcouto, line 57: add a test for this.

packages/ui-components/src/utils/date.helper.ts
```diff
@@ -102,6 +105,7 @@ export const isDateValid = (date: DateInput): boolean => {
 	return newDate(date).isValid();
 };
 export const calculateDate = (date: DateInput, amount: number = 0, unit: dayjs.ManipulateType = 'days'): dayjs.Dayjs => newDate(date).add(amount, unit);
+export const calculateQuarterDate = (date: DateInput, amount: number = 0): dayjs.Dayjs => newDate(date).add(amount, 'quarter');
```
djcouto, line 108: add a test for this.

djcouto (changes requested):
Few changes, but overall looks great to me.

Nicely done 👍

packages/ui-components/src/components/relative-time-picker/relative-time-picker.config.ts
GoncaloCamaz: Hehe it was a prop from the dropdown that was available to set this height but I can do it with in the scss file 👍

packages/ui-components/src/components/relative-time-picker/relative-time-picker.config.ts
GoncaloCamaz: Dependent on the comment above regarding the config file

GoncaloCamaz: Will look at this refactor after full implementation 👍

miguelpinto98 (changes requested):
👏  Good job miúdo! It's almost done!
I only left some suggestions that are easy to resolve.
My only concern is if we should pass with the timezone name, the offset value too. Think on that please.

packages/ui-components/src/components/relative-time-picker/relative-time-picker.helper.ts
```diff
@@ -0,0 +1,154 @@
...
+	const calculatedDate = newTimezoneDate(timeZone, option.endDate.date);
+
+	return buildSelectedRangeDetail(nowDateTime, calculatedDate, option.labelRangeFormatter);
+};
+
+export const buildRelativeAmountOfUnitsConfigRange = (option: IRelativeTimePickerOption, timeZone: string): SelectedRangeDetail => {
+	const nowDateTime = newTimezoneDate(timeZone);
+	const startDateConfig = option.startDate;
+	const calculatedDate =
+		startDateConfig.unit !== 'quarter'
```
djcouto, line 74: Move this quarter logic check to the `calculateDate` util.

packages/ui-components/src/components/time-picker-select-option/time-picker-select-option.scss
```diff
@@ -0,0 +1,61 @@
...
+		justify-content: space-between;
+		align-items: center;
+		min-width: 0;
+		width: 100%;
+
+		.item-label {
+			@include kv-font-span-regular;
+			transition: font-weight 100ms linear;
+			color: kv-color('neutral-2');
+			white-space: nowrap;
```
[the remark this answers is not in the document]
GoncaloCamaz: Will update 👍

packages/ui-components/src/components/relative-time-picker/relative-time-picker.config.ts
```diff
@@ -0,0 +1,434 @@
...
+			}
+		}
+	]
+];
+
+export const CUSTOMIZE_OPTION_HEIGHT = 41;
+export const TIMEZONE_DROPDOWN_HEIGHT = 41;
+export const SELECT_OPTION_HEIGHT = 32;
+export const GROUP_GAP = 12;
+export const MAX_HEIGHT = 347;
```
djcouto, line 434: All these CSS options should be passed to the stylesheet in order to be synchronized. This will avoid weird behaviours if they ever get changed and we forgot to update them.

djcouto (approved):
Left you some comments. 
Apart from that, LGTM 👍

packages/ui-components/src/components/relative-time-picker/relative-time-picker.types.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export type DateTimeCalculation = {
+	// Date which the calculation should be relative to.
+	// if undefined, now will be used
+	date?: DateInput;
+	// the amount of units to add to the relative time
+	// if this number is negative, the amount will be subtracted
+	amount?: number;
+	// the unit type to add/subtract to the relative time
+	unit?: string;
```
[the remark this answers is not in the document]
GoncaloCamaz: Because unit can also be a quarter and the dayjs Manipulate type excludes the quarter option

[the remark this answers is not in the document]
djcouto: You could make this type a union of those two, right?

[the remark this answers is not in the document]
GoncaloCamaz: Yep sure but that will not remove all the castings

packages/ui-components/src/components/input-wrapper/input-wrapper.scss
```diff
@@ -0,0 +1,42 @@
+@import '../../assets/styles/globals';
+
+:host {
+	display: block;
+}
+
+.input-container {
+	background-color: kv-color('neutral-7');
+	padding: 0 $spacing-4x;
+	height: 32px;
```
laraujo7, line 10: SHould be extract this height to a variable `input-height`

miguelpinto98 (approved):
🚀

pedrorebelocosta (approved):

PR #257, opened 2023-03-10
feat(dropdown): add input size configuration

GoncaloCamaz:
Add input size prop to change the size of the dropdown opening element.


pedrorebelocosta (approved):

packages/ui-components/src/components/dropdown/dropdown.tsx
```diff
@@ -55,6 +58,7 @@ export class KvDropdown implements IDropdown, IDropdownEvents {
 							<kv-text-field
 								{...this.getInputConfig()}
 								id="dropdown-input"
+								size={this.inputSize}
```
djcouto, line 61: this should be passed inside the `inputConfig` prop

packages/ui-components/src/components/multi-select-dropdown/multi-select-dropdown.tsx
```diff
@@ -170,7 +173,14 @@ export class KvMultiSelectDropdown implements IMultiSelectDropdown, IMultiSelect
 
 		return (
 			<Host>
-				<kv-dropdown inputConfig={this.getInputConfig()} isOpen={this.isOpen} onOpenStateChange={this.openStateChangeHandler} disabled={this.disabled} exportparts="input">
+				<kv-dropdown
+					inputConfig={this.getInputConfig()}
+					inputSize={this.inputSize}
```
djcouto, line 178: same here

packages/ui-components/src/components/single-select-dropdown/single-select-dropdown.tsx
```diff
@@ -52,6 +53,8 @@ export class KvSingleSelectDropdown implements ISingleSelectDropdown, ISingleSel
 	@Prop({ reflect: true }) minHeight?: string;
 	/** @inheritdoc */
 	@Prop({ reflect: true }) maxHeight?: string;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) inputSize?: EComponentSize = EComponentSize.Large;
```
djcouto, line 57: Same here

djcouto (changes requested):

packages/ui-components/src/components/single-select-dropdown/single-select-dropdown.tsx
GoncaloCamaz: Most of the props inside this component are used to configure the input. I believe it would be a great refactor here but merging this inside an inputConfig alongside all the other props related with the input would introduce major breaking changes both in this component and tests and in MAPS. Should we consider this refactor in another PR?

djcouto: Hum, you are correct, let's do this on another PR 😄

djcouto (approved):
:shipit:

PR #261, opened 2023-03-14
feat(description-list): add component

pedrorebelocosta:
Default:
image

Tooltip text:
image

Icon toggletip:
image


packages/ui-components/src/components/description-list/description-list.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+	render() {
+		return (
+			<Host>
+				<div class="description-list-container">
+					{this.items?.map(({ title, description, tooltipInfo }) => (
+						<Fragment>
+							<div class="title">{title}</div>
+							<div class="description">
+								<kv-tooltip text={tooltipInfo?.text && !tooltipInfo.icon ? tooltipInfo.text : ''} options={TEXT_TOOLTIP_CONFIG}>
```
djcouto, line 23: This seems very odd to me. Do we only want to show text if it exists and if there's no icon? 🤔

At least move this a util so it becomes clear what you want to do here.

Also, seems to me that this condition should apply to the entire `kv-tooltip`, not only the text prop.

djcouto: Also, seems to me that this tooltip config will need to be configurable via prop.

packages/ui-components/src/components/description-list/description-list.tsx
```diff
@@ -0,0 +1,38 @@
...
+									{description}
+								</kv-tooltip>
+								{tooltipInfo?.icon && (
```
djcouto, line 26: We can have a tooltip and a toggle tip in the same description item? I don't think so. 🤔

djcouto (changes requested):
Left a few comments.
Furthermore, don't forget to export the angular module for this new component.

packages/ui-components/src/components/description-list/description-list.config.ts
```diff
@@ -0,0 +1,7 @@
+import { ComputePositionConfig, offset } from '@floating-ui/dom';
+import { ETooltipPosition } from '../../types';
+
+export const TEXT_TOOLTIP_CONFIG: Partial<ComputePositionConfig> = {
+	placement: ETooltipPosition.Bottom,
+	middleware: [offset(4)]
```
GoncaloCamaz, line 6: Do you think we should have this 4 in a config to be more understanding what kind of offset we are applying here? Top/bottom?

packages/ui-components/src/components/description-list/description-list.tsx
GoncaloCamaz: In case we have both, should we consider having two different config properties like:
- tooltipInfo
- toggletipInfo 

This to avoid having properties with a name of one component being used in another, ex: "<kv-toggle-tip text={tooltipInfo.text}"

GoncaloCamaz (changes requested):
Nice work! Just some small questions 👍

packages/ui-components/src/components/description-list/description-list.tsx
pedrorebelocosta: > This seems very odd to me. Do we only want to show text if it exists and if there's no icon? 🤔
That is indeed the intended behaviour, if there's text but no icon provided, show it on the tooltip.

> At least move this a util so it becomes clear what you want to do here.
I will this verification to a helper/util method so it becomes clearer.

> Also, seems to me that this condition should apply to the entire `kv-tooltip`, not only the text prop.
My bad but I can't seem to understand what you mean by this, are we talking about conditionally rendering this tooltip?

packages/ui-components/src/components/description-list/description-list.tsx
pedrorebelocosta: > We can have a tooltip and a toggle tip in the same description item? I don't think so. 🤔

Only one or the other

djcouto (approved):
LGTM

GoncaloCamaz (approved):

PR #265, opened 2023-03-21
fix(modal): enable style customization on modal component

djcouto:


pedrorebelocosta (approved):

pedrorebelocosta (approved):

pedrorebelocosta:
I think this should be a breaking change as the interface had `closeOnOverlayClick` required

GoncaloCamaz (approved):

PR #266, opened 2023-03-22
fix(copy-to-clipboard): change component behavior on hover and minor updates

miguelpinto98:


pedrorebelocosta (approved):

packages/ui-components/src/components/copy-to-clipboard/copy-to-clipboard.tsx
```diff
@@ -39,21 +39,19 @@ export class KvCopyToClipboard implements ICopyToClipboard {
...
+				<kv-tooltip text={this.tooltipText} options={TOOLTIP_CONFIG} delay={TOOLTIP_DELAY} onClick={this.onTextCopy}>
+					<div part="content" class="copy-to-clipboard-container">
+						<slot></slot>
+						<div
+							class={{
+								'state-icon': true,
+								'state-icon--success': this.copyState === ECopyToClipboardState.Copied
+							}}
+						>
+							<kv-icon name={ICON_CONFIGS[this.copyState]} />
```
djcouto, line 51: We can either export this icon part to be styled from outside.

PR #274, opened 2023-04-01
fix(schema-form-checkbox-widget): fix toogle selected options

laraujo7:


feliciofilipe (approved):
LGTM :muscle:

GoncaloCamaz (approved):

packages/react-ui-components/src/components/SchemaForm/Widgets/CheckboxesWidget/utils.ts
```diff
@@ -77,6 +77,10 @@ export const toggleSelectedOptions = <T extends CheckboxOption>(
 	const allButtonsSelected = selectedOptions.length === allOptions.length;
 
 	if (allButton) {
+		if (selectedOptionValue !== ALL_BUTTON_VALUE && allButtonsSelected) {
```
djcouto, line 80: We can move this to below the `(selectedOptionValue === ALL_BUTTON_VALUE)` condition just like this:
```typescript
if (allButtonsSelected) {
    return [selectedOptionValue];
}
```

packages/react-ui-components/src/components/SchemaForm/Widgets/CheckboxesWidget/utils.ts
```diff
@@ -90,6 +94,10 @@ export const toggleSelectedOptions = <T extends CheckboxOption>(
 
 			return allOptions.map(({ value }) => value);
 		}
+
+		if (selectedOptions.includes(selectedOptionValue) && selectedOptions.length === minItems) {
```
djcouto, line 98: Can you explain what the purpose of this is change? It might not be necessary 🤔 
From my understanding, before de-selecting an option, we must ensure the selected options' minimum constraint is met. Due to this, if the selected options count is already at the minimum, we must not perform any action and leave it as it is.
This is what is being done between #L104 and #L108.

djcouto (changes requested):

djcouto (approved):
🚀

PR #275, opened 2023-04-03
fix(code-editor): add kelvin theme to match DS

pedrorebelocosta:


GoncaloCamaz (approved):

djcouto (changes requested):
Take a look at the `setupEditor`. I think you are always setting the theme to custom regardless of the theme prop value.

PR #277, opened 2023-04-04
refactor(code-editor)!: separate font loading logic and add kelvin theme

pedrorebelocosta:
BREAKING CHANGE: `readOnly` and `paddingTop` properties were moved to the `customOptions` object

### DIFF
--- a/packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
+++ b/packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
@@ -1,53 +1,45 @@
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo } from 'react';
 import Editor, { useMonaco } from '@monaco-editor/react';
-import { DEFAULT_EDITOR_LANGUAGE, DEFAULT_EDITOR_THEME, DEFAULT_PADDING_TOP, FONT_NOT_FOUND_ERROR } from './config';
-import { ECodeEditorTheme, ICodeEditorProps, OnEditorChangeCallback } from './types';
+import { DEFAULT_CODE_EDITOR_LANGUAGE, DEFAULT_CODE_EDITOR_THEME, KELVIN_CODE_EDITOR_THEME } from './config';
+import { ECodeEditorTheme, ICodeEditorProps, OnCodeEditorChange } from './types';
 import { KvLoader } from '../stencil-generated';
-import { getEditorOptions } from './utils';
+import { getEditorOptions, getFontOptions } from './utils';
+import { useFontsApi } from '../../utils';
+
+const CodeEditorLoader = () => <KvLoader isLoading />;
 
 export const KvCodeEditor = ({
 	code,
-	readOnly = false,
-	loadingComponent = <KvLoader isLoading />,
-	language = DEFAULT_EDITOR_LANGUAGE,
-	theme = DEFAULT_EDITOR_THEME,
-	customTheme,
-	paddingTop = DEFAULT_PADDING_TOP,
+	language = DEFAULT_CODE_EDITOR_LANGUAGE,
+	theme = DEFAULT_CODE_EDITOR_THEME,
+	customTheme = KELVIN_CODE_EDITOR_THEME,
+	customOptions,
+	LoadingComponent = CodeEditorLoader,
 	onChange
 }: ICodeEditorProps) => {
 	const monaco = useMonaco();
-	const [isFontLoaded, setFontLoaded] = useState(false);
+	const options = useMemo(() => getEditorOptions(customOptions), [customOptions]);
+	const { isFontLoaded, loadFont } = useFontsApi(getFontOptions(options));
 
-	const options = useMemo(() => getEditorOptions({ readOnly, padding: { top: paddingTop } }), [readOnly, paddingTop]);
+	const hasLoaded = useMemo(() => monaco && isFontLoaded, [monaco, isFontLoaded]);
 
-	const loadFont = useCallback(async () => {
-		try {
-			const { fontSize, fontFamily } = options;
-			await document.fonts.load(`${fontSize}px ${fontFamily}`);
-			setFontLoaded(true);
-		} catch (error) {
-			throw new Error(FONT_NOT_FOUND_ERROR);
-		}
-	}, [options.fontSize, options.fontFamily]);
-
-	const setupEditor = useCallback(() => {
+	const defineCustomTheme = useCallback(() => {
 		if (!monaco || !customTheme) return;
 		monaco.editor.defineTheme(ECodeEditorTheme.Custom, customTheme);
-		monaco.editor.setTheme(ECodeEditorTheme.Custom);
-	}, [monaco, customTheme]);
+	}, [customTheme, monaco]);
 
-	const onTextChange: OnEditorChangeCallback = useCallback(value => onChange?.(value), [onChange]);
+	const onTextChange: OnCodeEditorChange = useCallback(value => onChange?.(value), [onChange]);
 
 	useEffect(() => {
 		loadFont();
-		setupEditor();
-	}, [loadFont, setupEditor]);
+		defineCustomTheme();
+	}, [loadFont, defineCustomTheme]);
 
-	if (!isFontLoaded) {
-		return loadingComponent;
+	if (!hasLoaded) {
+		return <LoadingComponent />;
 	}
 
-	return <Editor language={language} theme={theme} options={options} value={code} loading={loadingComponent} onChange={onTextChange} />;
+	return <Editor language={language} theme={theme} options={options} loading={<LoadingComponent />} value={code} onChange={onTextChange} />;
 };
 
 export default KvCodeEditor;
--- a/packages/react-ui-components/src/components/CodeEditor/types.ts
+++ b/packages/react-ui-components/src/components/CodeEditor/types.ts
@@ -1,8 +1,9 @@
 import type { editor } from 'monaco-editor';
+import { FunctionComponent } from 'react';
 
-export type OnEditorChangeCallback = (value: string | undefined) => void;
-export type CustomThemeData = editor.IStandaloneThemeData;
-export type CustomEditorOptions = editor.IEditorOptions & editor.IGlobalEditorOptions;
+export type OnCodeEditorChange = (value: string | undefined) => void;
+export type CodeEditorOptions = editor.IGlobalEditorOptions & editor.IEditorOptions;
+export type CodeEditorTheme = editor.IStandaloneThemeData;
 
 export enum ECodeEditorTheme {
 	Dark = 'vs-dark',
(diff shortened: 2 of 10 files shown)


packages/react-ui-components/src/utils/useFont/useFontsApi.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const FONT_NOT_FOUND_ERROR = `Font was not found, make sure the font is included in your css`;
+
+export const useFontsApi = (fontFamily: string, fontSize: number) => {
+	const fontName = useMemo(() => `${fontSize}px ${fontFamily}`, [fontFamily, fontSize]);
+
+	const checkFont = useCallback(() => document.fonts.check(fontName), [fontName]);
+
+	const loadFont = useCallback(async () => {
+		if (checkFont()) return;
```
djcouto, line 11: Shouldn't it be? `if (!checkFont()) return;`?

djcouto (changes requested):
Nice improvement 👍

packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -1,53 +1,42 @@
...
+import { DEFAULT_EDITOR_LANGUAGE, DEFAULT_EDITOR_THEME, DEFAULT_PADDING_TOP, KELVIN_CODE_EDITOR_THEME } from './config';
 import { ECodeEditorTheme, ICodeEditorProps, OnEditorChangeCallback } from './types';
 import { KvLoader } from '../stencil-generated';
 import { getEditorOptions } from './utils';
+import { useFontsApi } from '../../utils/useFont/useFontsApi';
 
 export const KvCodeEditor = ({
 	code,
 	readOnly = false,
 	loadingComponent = <KvLoader isLoading />,
```
djcouto, line 12: This component will be recreated on every mount which is not ideal.
Can we create a "CodeEditorLoader" component and pass the reference to the component instead of react node?
Something like this:
```typescript
export const KvCodeEditor = ({
    ...
    LoadingComponent = CodeEditorLoader,
	...
}: ICodeEditorProps) => {
	...
	if (!checkFont()) {
		return <LoaderComponent />;
	}
    ...
};
```
Also, component props are usually capitalised to distinguish them from other props types.

packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -1,53 +1,42 @@
...
 
 	useEffect(() => {
 		loadFont();
-		setupEditor();
-	}, [loadFont, setupEditor]);
+		loadCustomTheme();
+	}, [loadFont, loadCustomTheme]);
 
-	if (!isFontLoaded) {
+	if (!checkFont()) {
```
djcouto, line 35: Should you also check here for the "monaco" to be available and for the theme to be loaded?

djcouto (changes requested):
Sorry, I needed to add to more comments to my previous review 😄

packages/react-ui-components/src/utils/useFont/useFontsApi.ts
pedrorebelocosta: `checkFont` verifies if the font has already been loaded, if it returns `true`, there's no need for this function's execution to continue

packages/react-ui-components/src/utils/useFontsApi.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const FONT_NOT_FOUND_ERROR = `Font was not found, make sure the font is included in your css`;
+
+export const useFontsApi = (fontFamily: string, fontSize: number) => {
+	const fontName = useMemo(() => `${fontSize}px ${fontFamily}`, [fontFamily, fontSize]);
+
+	const checkFont = useCallback(() => document.fonts.check(fontName), [fontName]);
+
+	const loadFont = useCallback(async () => {
+		if (checkFont()) return;
```
djcouto, line 11: What about a `isFontAvailable` memo instead of a callback? Seems better to me.
Also, add a return type to this hook.

```typescript
export const useFontsApi = (fontFamily: string, fontSize: number) => {
    const fontName = useMemo(() => `${fontSize}px ${fontFamily}`, [fontFamily, fontSize]);
    const isFontAvailable = useMemo(() => document.fonts.check(fontName), [fontName]);

	const checkFont = useCallback(() => document.fonts.check(fontName), [fontName]);
	const loadFont = useCallback(async () => {
		if (isFontAvailable) return;

		try {
			await document.fonts.load(fontName);
		} catch (error) {
			throw new Error(FONT_NOT_FOUND_ERROR);
		}
	}, [fontName]);

	return { checkFont, loadFont, isFontAvailable };
};
```

packages/react-ui-components/src/utils/useFontsApi.ts
```diff
@@ -0,0 +1,21 @@
+import { useCallback, useMemo } from 'react';
+
+const FONT_NOT_FOUND_ERROR = `Font was not found, make sure the font is included in your css`;
+
+export const useFontsApi = (fontFamily: string, fontSize: number) => {
+	const fontName = useMemo(() => `${fontSize}px ${fontFamily}`, [fontFamily, fontSize]);
```
miguelpinto98, line 6: Since we are loading these properties from a callback, we should not memo this. We should build this string inside the loadFont.

packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -1,53 +1,45 @@
...
+	}, [customTheme, monaco]);
 
 	const onTextChange: OnEditorChangeCallback = useCallback(value => onChange?.(value), [onChange]);
 
 	useEffect(() => {
 		loadFont();
-		setupEditor();
-	}, [loadFont, setupEditor]);
+		defineCustomTheme();
+	}, [loadFont, defineCustomTheme]);
```
miguelpinto98, line 36: probably we need to add monaco to the dependency list, to reload the data if the instance change.

djcouto (changes requested):

packages/react-ui-components/src/components/CodeEditor/types.ts
```diff
@@ -1,8 +1,29 @@
...
 export type OnEditorChangeCallback = (value: string | undefined) => void;
 export type CustomThemeData = editor.IStandaloneThemeData;
-export type CustomEditorOptions = editor.IEditorOptions & editor.IGlobalEditorOptions;
+
+export type KvEditorPaddingOptions = {
+	top?: number;
+	bottom?: number;
+};
+
+export type KvEditorOptions = {
```
miguelpinto98, line 12: we are creating a limitation on editor options to only allow passing these properties. 
We should extend all editor options here. In rules/process map editor we are using directly the monaco-editor, and if we refactor these components to use this component we should allow the other monaco options here, because they use other options.

miguelpinto98 (changes requested):
👍 Some changes here are for good! We are marking this PR with a breaking change, can you describe what changes? And what should be done? (IDK if this PR should be breaking)

packages/react-ui-components/src/components/CodeEditor/types.ts
```diff
@@ -1,8 +1,29 @@
...
+	/** Use this property to set the editor to a read only mode where the user cannot modify the value. */
+	readOnly?: boolean;
+	/** Use this property to add some padding (in px) at the top of the editor */
+	padding?: KvEditorPaddingOptions;
+};
+
+type ExcludedEditorOptions = 'fontSize' | 'fontFamily';
```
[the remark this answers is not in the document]
pedrorebelocosta: Since the type defined by monaco is `string | undefined` for `fontFamily` and `number | undefined` for `fontSize` I've decided to override these props to avoid undefined checks when passing these props around (for instance, to `useFontsApi`), if you disagree on this approach, let me know

packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
pedrorebelocosta: `monaco` is already on the dependency list of `defineCustomTheme` and `hasLoaded`, should we include it here?

packages/react-ui-components/src/components/CodeEditor/types.ts
pedrorebelocosta: I'm torn between these 2 approaches, exposing everything monaco has vs having this intermediate type that allows us to have some abstraction over what editor we use underneath.

@djcouto what's your take?

packages/react-ui-components/src/utils/useFontsApi.ts
pedrorebelocosta: looks good to me :D

packages/react-ui-components/src/components/CodeEditor/types.ts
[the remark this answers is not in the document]
miguelpinto98: if we turn all monaco options available from component we can't require the user to set a fontFamily or size.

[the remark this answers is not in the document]
pedrorebelocosta: I'm merging our defaults with the `customOptions` the parent provides, so these 2 properties will always be defined.

Depending on the outcome of [this](https://github.com/kelvininc/ui-components/pull/277#discussion_r1157269263) conversation the approach here must be different.

packages/react-ui-components/src/components/CodeEditor/types.ts
djcouto: IMO, I'm a fan of keeping this library agnostic. Today we are using monaco but it can run unmaintained tomorrow or a new cutting-edge library can come up. With that being said, these settings are a bit monaco-specific, so I would say to go with @miguelpinto98's suggestion and expose everything monaco has.

packages/react-ui-components/src/components/CodeEditor/types.ts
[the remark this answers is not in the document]
pedrorebelocosta: I'm going to change this to expose all props as that seems to be the consensus, thanks!

packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -1,53 +1,45 @@
...
-	readOnly = false,
-	loadingComponent = <KvLoader isLoading />,
-	language = DEFAULT_EDITOR_LANGUAGE,
-	theme = DEFAULT_EDITOR_THEME,
-	customTheme,
-	paddingTop = DEFAULT_PADDING_TOP,
+	language = DEFAULT_CODE_EDITOR_LANGUAGE,
+	theme = DEFAULT_CODE_EDITOR_THEME,
+	customTheme = KELVIN_CODE_EDITOR_THEME,
+	customOptions = DEFAULT_CODE_EDITOR_OPTIONS,
```
djcouto, line 16: Not sure if we should merge the provided custom options with the default ones. Since this object can be a bit extensive, probably we should not force the user to write every option and instead only define the ones he wants to override. 🤔

djcouto (approved):

packages/react-ui-components/src/components/CodeEditor/CodeEditor.tsx
pedrorebelocosta: my bad, left it there after making the changes that were discussed [here](https://github.com/kelvininc/ui-components/pull/277#discussion_r1158574157)

GoncaloCamaz (approved):

miguelpinto98 (approved):
👍

PR #280, opened 2023-04-11
feat(modal): add topbar border css custom property

laraujo7:


packages/ui-components/src/components/modal/modal.base.scss
```diff
@@ -59,7 +60,7 @@
 		justify-content: space-between;
 		padding: 0 $spacing-4x;
 		height: var(--modal-topbar-height);
-		border-bottom: 1px solid kv-color('neutral-7');
+		border-bottom: var(--modal-topbar-border) solid kv-color('neutral-7');
```
djcouto, line 63: Is this really needed? The topbar part is exported and can be easily configured 😄
Nevertheless, if you choose to continue with this PR, change the CSS custom variable to configure the entire border-bottom and not only its width.

GoncaloCamaz (approved):

packages/ui-components/src/components/modal/modal.base.scss
laraujo7: Yap, I didn't realize that we had a part for the topbar. Thanks, I'm going to close this one :muscle:

PR #281, opened 2023-04-11
feat(time-range): add new time range picker component

GoncaloCamaz:
Screenshot 2023-04-11 at 12 20 00


packages/ui-components/src/components/absolute-time-picker/absolute-time-picker.tsx
```diff
@@ -48,24 +49,39 @@ export class KvAbsoluteTimePicker implements IAbsoluteTimePicker, IAbsoluteTimeP
...
 
 	@Watch('selectedRangeDates')
 	handleSelectedRangeDatesChange(value: string[] = []) {
-		// Reset the component state if the selected dates change
 		if (value.length === 0) {
-			this.fromInputValue = '';
-			this.toInputValue = '';
-			this.minDate = '';
-			this.maxDate = '';
+			this.setFromToInputValues('', '');
```
djcouto, line 74: I see this a lot in this file. Please choose one of following the approaches:
- Change the `setFromToInputValues` method arguments to be optional and default to an empty string
- Create a method named `resetInputValues` that abstract this trick and give it a name.

packages/ui-components/src/components/absolute-time-picker/absolute-time-picker.tsx
```diff
@@ -74,23 +90,31 @@ export class KvAbsoluteTimePicker implements IAbsoluteTimePicker, IAbsoluteTimeP
...
-		this.toInputValue = newValue.to;
+		this.setFromToInputValues(newValue.from, newValue.to);
 		const [from] = this.selectedRangeDates;
 		const date = newDate(from);
 		if (date.isValid()) {
 			this.displayedMonth = date;
 		}
 	}
 
+	private setMinMaxDateValues = (min: string, max: string) => {
```
djcouto, line 108: Does `setDateLimits` naming seem better to you?

packages/ui-components/src/components/absolute-time-picker/absolute-time-picker.tsx
```diff
@@ -74,23 +90,31 @@ export class KvAbsoluteTimePicker implements IAbsoluteTimePicker, IAbsoluteTimeP
...
+		this.minDate = min;
+		this.maxDate = max;
+	};
+
+	private setFromToInputValues = (from: string, to: string) => {
```
djcouto, line 113: Does `setInputsValues` naming seem better to you?

packages/ui-components/src/components/date-time-input/date-time-input.tsx
```diff
@@ -150,6 +155,7 @@ export class DateTimeInput implements IDateTimeInput, IDateTimeInputEvents {
 								ref={input => (this.nativeInput = input as HTMLInputElement)}
 								type={EInputFieldType.Text}
 								name={this.inputName}
+								disabled={this.disabled}
```
djcouto, line 158: Should the input be focusable if it's disabled?

packages/ui-components/src/components/date-time-input/date-time-input.tsx
```diff
@@ -137,7 +142,7 @@ export class DateTimeInput implements IDateTimeInput, IDateTimeInputEvents {
 		return (
 			<Host>
 				<div class="date-time-input-container">
-					<kv-form-label label={this.label} />
+					<kv-form-label label={this.label} required={this.required} />
```
djcouto, line 145: Can it be required if it's disabled?

packages/ui-components/src/components/date-time-input/test/date-time-input.e2e.ts
```diff
@@ -0,0 +1,61 @@
...
+			let dateTimeInputComponent: E2EElement;
+
+			beforeEach(async () => {
+				dateTimeInputComponent = await page.find('kv-date-time-input');
+				spyChangeEvent = await dateTimeInputComponent.spyOnEvent('textChange');
+
+				const dateTimeInputElement = await page.find('input');
+				await dateTimeInputElement.type('Teste');
+
+				await page.waitForTimeout(2);
```
djcouto, line 23: This is deprecated. We should not wait for a specific period of time.
Instead, we should wait for an event to be dispatched. See if this works:

```typescript
await page.waitForEvent('textChange');
```

packages/ui-components/src/components/relative-time-picker/relative-time-picker.helper.ts
```diff
@@ -62,6 +62,11 @@ export const buildRelativeAmountOfUnitsConfigRange = (option: IRelativeTimePicke
 	const nowDateTime = newTimezoneDate(timeZone);
 	const { amount, unit } = option.startDate;
 	const calculatedDate = calculateDate(nowDateTime, amount, unit);
+
+	if (amount === 0) {
```
djcouto, line 66: 🔨

packages/ui-components/src/components/text-field/text-field.utils.ts
```diff
@@ -10,13 +10,7 @@ export function getInputMaskConfig(type: EInputFieldType): Inputmask.Options {
...
 	if (type === EInputFieldType.DateTime) {
 		return {
-			alias: 'datetime',
-			inputFormat: 'dd-mm-yyyy HH:MM:ss',
-			displayFormat: 'dd-mm-yyyy 00:00:00',
-			placeholder: 'dd-mm-yyyy 00:00:00',
-			jitMasking: true,
-			showMaskOnHover: false,
-			clearMaskOnLostFocus: false
+			...DATE_TIME_INPUT_MASK_CONFIG
```
djcouto, line 13: Do we need the destructuring here? Can't we just return the config constant?

packages/ui-components/src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,461 @@
...
+	/** @inheritdoc */
+	@Prop({ reflect: true }) relativeTimePickerOptions?: IRelativeTimePickerOption[][] = DEFAULT_RELATIVE_TIME_OPTIONS_GROUPS;
+	/** @inheritdoc */
+	@Prop({ reflect: false }) timezones?: string[] = getTimezonesNames();
+	/** @inheritdoc */
+	@Prop({ reflect: true }) disableTimezoneSelection?: boolean = false;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) displayCustomizeInterval?: boolean = true;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) displayTimezoneDropdown?: boolean = true;
```
djcouto, line 59: isn't this the same as setting timezones to empty array? 🤔

packages/ui-components/src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,461 @@
...
+
+		const defaultTimezone = getDefaultTimezone();
+		if (this.timezones.includes(defaultTimezone)) {
+			return {
+				name: defaultTimezone,
+				offset: getTimezoneOffset(defaultTimezone)
+			};
+		}
+
+		return undefined;
```
djcouto, line 121: This is unnecessary

packages/ui-components/src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,461 @@
...
+		}
+
+		if (!isEmpty(this.selectedTimeOption) && isEmpty(this.selectedTimeState)) {
+			this.selectedTimeState = this.selectedTimeOption;
+		}
+	}
+
+	private getSelectedTimezone = (): ITimePickerTimezone | undefined => {
+		if (this.selectedTimeState !== undefined) {
+			return this.selectedTimeState?.timezone;
```
djcouto, line 110: No need for optional chaining.

packages/ui-components/src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,461 @@
...
+			return {
+				name: defaultTimezone,
+				offset: getTimezoneOffset(defaultTimezone)
+			};
+		}
+
+		return undefined;
+	};
+
+	private onDropdownChange = (event: CustomEvent<boolean>) => {
```
djcouto, line 124: Let's destruct this event and give it some proper naming.

```suggestion
	private onDropdownChange = ({ detail: isDropdownOpen }: CustomEvent<boolean>) => {
```

packages/ui-components/src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,461 @@
...
+			this.showCalendar = !currentValue;
+		}
+	};
+
+	private getRelativeTimeInputConfig(): IRelativeTimeInput | undefined {
+		if (this.selectedTimeState?.key !== CUSTOMIZE_INTERVAL_KEY) {
+			return getRelativeTimeInputText(this.relativeTimePickerOptions, this.selectedTimeState, this.getSelectedTimezone().name);
+		}
+
+		return undefined;
```
djcouto, line 253: Unnecessary return statement

djcouto (changes requested):
Few comments, great job 👍

packages/ui-components/src/components/date-time-input/date-time-input.tsx
GoncaloCamaz: I assumed that this can be required even though is disabled because imagine a use case where the disable is only temporary and its changed after the user does a specific action. Even though he cannot update this input, it continues to be a required field.

packages/ui-components/src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,461 @@
...
+	componentWillLoad() {
+		if (isEmpty(this.selectedTimeOption)) {
+			this.selectedTimeState = {
+				key: DEFAULT_SELECTED_TIME_KEY,
+				range: getLast24HoursRange(),
+				timezone: this.getSelectedTimezone()
+			};
+
+			return;
+		}
```
miguelpinto98, line 101: else and remove the return? we are re-checking `selectedTimeOption`

packages/ui-components/src/components/time-picker/time-picker.tsx
miguelpinto98: and try to apply the same approach on the following functions

packages/ui-components/src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,461 @@
...
+			timezone: this.getSelectedTimezone()
+		};
+		this.calendarViewLocked = false;
+	};
+
+	private onClickSeeCustomInterval = (event: CustomEvent<string>) => {
+		const key = event.detail;
+		if (this.selectedTimeState && this.selectedTimeState.key !== CUSTOMIZE_INTERVAL_KEY) {
+			this.selectedTimeState = {
+				key: key,
```
miguelpinto98, line 159: key,

packages/ui-components/src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,461 @@
...
+			this.timePickerView = ETimePickerView.AbsoluteTimePicker;
+		} else {
+			this.calendarViewLocked = true;
+		}
+	};
+
+	private onSelectedTimezoneChange = (event: CustomEvent<ITimePickerTimezone>) => {
+		const timezone = event.detail;
+		const previousTimezone = this.selectedTimeState.timezone.name;
+		if (this.selectedTimeState.key === CUSTOMIZE_INTERVAL_KEY) {
```
miguelpinto98, line 175: can we simplify this?

```
const range = this.selectedTimeState.key === CUSTOMIZE_INTERVAL_KEY
	? getTimestampFromDateRange(this.selectedTimeState.range, previousTimezone, timezone.name)
	: undefined;

this.selectedTimeState = {
	...this.selectedTimeState,
	range,
	timezone
};
```

packages/ui-components/src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,461 @@
...
+				timezone
+			};
+		}
+	};
+
+	private onClickApply = () => {
+		const payload = {
+			key: this.selectedTimeState.key,
+			range: this.selectedTimeState.range,
+			timezone: this.getSelectedTimezone()
```
miguelpinto98, line 193: the timezone value that we have on `selectedTimeState` can differ?

packages/ui-components/src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,461 @@
...
+			this.selectedTimeState = {
+				key: DEFAULT_SELECTED_TIME_KEY,
+				range: getLast24HoursRange(),
+				timezone: this.getSelectedTimezone()
+			};
+		}
+		this.timePickerView = ETimePickerView.RelativeTimePicker;
+	};
+
+	private onClickCancel = (event: CustomEvent<MouseEvent>) => {
```
miguelpinto98, line 221: this function and `onClickBack` have a lot in common, can we reduce the replication?

packages/ui-components/src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,461 @@
...
+		if (this.timePickerView !== ETimePickerView.FullView) {
+			this.timePickerView = ETimePickerView.RelativeTimePicker;
+		}
+	};
+
+	private onShowCalendarClick = () => {
+		if (!this.calendarViewLocked) {
+			const currentValue = this.showCalendar;
+			this.timePickerView = currentValue ? ETimePickerView.RelativeTimePicker : ETimePickerView.FullView;
+			this.showCalendar = !currentValue;
```
miguelpinto98, line 244: currentValue is necessary?

miguelpinto98 (changes requested):
Almost done! Some questions to discuss were added.

packages/ui-components/src/components/relative-time-picker/relative-time-picker.helper.ts
GoncaloCamaz: This was a really specific case that only happens when we are comparing a relative date to now and the amount of units we are removing is 0. For example, considering "Today" example, we want the time from the start of the day until the current timestamp. Considering the config file: 

	`label:` `'Today',`
	`value:` `'today',`
	`comparisonConfig:` `ERelativeTimeComparisonConfig.RelativeAmountOfUnits,`
	`startDate:` {
		`amount:` `0,`
		`unit:` `'days'`
	`}`

We are using the `calculateDate` to add amounts of units to the current date. Therefore, when the amount is negative, we take x units from the current timestamp. But here in this specific case, we are removing 0 days from the current timestamp. That being said, we are actually asking for the current timestamp which will be a bad result because the interval will be from now until now. 

My concern here with changing the calculate date function is that we will be inserting unexpected behaviour on this util. We can see it both ways and maybe the suggestion that Diogo gave the other day would be a better alternative for this case that would consist in adding in the config file maybe an enum to determine if we want the start of or the end of unit like the example bellow:

	`startDate:` {
		`amount:` `0,`
		`unit:` `'days'`
                `limit`: `EUnitRange.StartOf` || `EUnitRange.EndOf`
	`}`

But this change to the config file would only make sense to cover this edge case since this does change does not apply to the other `comparisonConfig` options.

packages/ui-components/src/components/absolute-time-picker/absolute-time-picker.scss
```diff
@@ -16,7 +16,8 @@
 		width: fit-content;
 
 		kv-icon {
-			--icon-color: #{kv-color('neutral-4')};}
+			--icon-color: #{kv-color('neutral-4')};
```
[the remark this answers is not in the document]
GoncaloCamaz: I believe this was introduced by the linter in the fix process but in terms of readability, I updated because it made more sense having the closing bracket bellow the kv-icon tag. I can rever it though but personally I think this change made more sense since we are following this approach everywhere.

packages/ui-components/src/components/relative-time-picker/relative-time-picker.tsx
```diff
@@ -60,22 +64,22 @@ export class KvRelativeTimePicker implements IRelativeTimePicker, IRelativeTimeP
...
+	@State() selectedOptionRange: SelectedTimestampRange;
 
 	/** @inheritdoc */
 	@Event() selectedRelativeTimeChange: EventEmitter<ITimePickerRelativeTime>;
 	/** @inheritdoc */
 	@Event() customizeIntervalClicked: EventEmitter<string>;
 	/** @inheritdoc */
 	@Event() timezoneChange: EventEmitter<ITimePickerTimezone>;
+	/** @inheritdoc */
+	@Event() inputWrapperClicked: EventEmitter<boolean>;
```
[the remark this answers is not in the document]
GoncaloCamaz: I agree! Maybe timezoneInputClicked would be more suitable.

packages/ui-components/src/components/time-picker/time-picker.tsx
GoncaloCamaz: Not needed 👍

packages/ui-components/src/components/date-time-input/date-time-input.tsx
djcouto: Fine by me then 👍

packages/ui-components/src/components/time-picker/time-picker.tsx
djcouto: It is, coherence is always a good policy 👍

packages/ui-components/src/components/input-wrapper/input-wrapper.tsx
```diff
@@ -19,16 +21,22 @@ export class KvInputWrapper implements IInputWrapper, IInputWrapperEvents {
...
-				<div class={{ 'input-container': true, 'input-container--content-hidden': !this.contentVisible }} onClick={this.handleContentClick}>
-					{this.contentVisible ? (
+				<div
+					class={{
+						'input-container': true,
+						'input-container--content-hidden': !this.contentVisible && !this.contentDisabled
+					}}
+					onClick={this.handleContentClick}
+				>
+					{this.contentVisible && !this.contentDisabled ? (
```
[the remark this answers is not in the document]
GoncaloCamaz: If the content should is blocked, it should not be visible, but will rename the prop contentDisabled to a more suitable name

packages/ui-components/src/components/absolute-time-picker/absolute-time-picker.scss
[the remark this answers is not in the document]
miguelpinto98: yes, I miss that part :(

packages/ui-components/src/components/date-time-input/test/date-time-input.e2e.ts
GoncaloCamaz: It did not work, but waitForChanges did!

packages/ui-components/src/components/date-time-input/date-time-input.tsx
GoncaloCamaz: Changed bellow to prevent this to be focusable if disabled

packages/ui-components/src/components/absolute-time-picker/absolute-time-picker.tsx
```diff
@@ -74,23 +90,41 @@ export class KvAbsoluteTimePicker implements IAbsoluteTimePicker, IAbsoluteTimeP
...
 			this.displayedMonth = date;
 		}
 	}
 
+	private setDateLimits = (min: string, max: string) => {
+		this.minDate = min;
+		this.maxDate = max;
+	};
+
+	private setInputValues = (from: string, to: string) => {
```
djcouto, line 113: Should these arguments be optional and default to empty string? Just to avoid having this:
```typescript
this.setInputValues('', ''); // this.setInputValues();
this.setInputValues(inputDate.startOf('day').format(DATETIME_INPUT_MASK), ''); // this.setInputValues(inputDate.startOf('day').format(DATETIME_INPUT_MASK));
```

packages/ui-components/src/components/absolute-time-picker/absolute-time-picker.tsx
```diff
@@ -189,8 +213,7 @@ export class KvAbsoluteTimePicker implements IAbsoluteTimePicker, IAbsoluteTimeP
 	};
 
 	private handleInputReset = () => {
-		this.fromInputValue = '';
-		this.toInputValue = '';
+		this.setInputValues('', '');
```
djcouto, line 216:
```suggestion
		this.resetInputValues();
```

packages/ui-components/src/components/absolute-time-picker/absolute-time-picker.tsx
```diff
@@ -248,8 +271,7 @@ export class KvAbsoluteTimePicker implements IAbsoluteTimePicker, IAbsoluteTimeP
 				this.displayedMonth = parsedDate;
 				if (!isEmpty(this.toInputValue)) {
 					this.emitSelectRangeDatesChangeEvent(parsedDateTime, dayjs(this.toInputValue, DATETIME_INPUT_MASK));
-					this.minDate = '';
-					this.maxDate = '';
+					this.setDateLimits('', '');
```
djcouto, line 274:
```suggestion
		this.resetDateLimits();
```

packages/ui-components/src/components/absolute-time-picker/absolute-time-picker.tsx
```diff
@@ -260,8 +282,7 @@ export class KvAbsoluteTimePicker implements IAbsoluteTimePicker, IAbsoluteTimeP
 				this.displayedMonth = parsedDate.subtract(1, 'month');
 				if (!isEmpty(this.fromInputValue)) {
 					this.emitSelectRangeDatesChangeEvent(dayjs(this.fromInputValue, DATETIME_INPUT_MASK), parsedDateTime);
-					this.minDate = '';
-					this.maxDate = '';
+					this.setDateLimits('', '');
```
djcouto, line 285:
```suggestion
		this.resetDateLimits();
```

packages/ui-components/src/components/date-time-input/test/date-time-input.e2e.ts
```diff
@@ -0,0 +1,61 @@
...
+			let dateTimeInputComponent: E2EElement;
+
+			beforeEach(async () => {
+				dateTimeInputComponent = await page.find('kv-date-time-input');
+				spyChangeEvent = await dateTimeInputComponent.spyOnEvent('textChange');
+
+				const dateTimeInputElement = await page.find('input');
+				await dateTimeInputElement.type('Teste');
+
+				await page.waitForChanges();
```
djcouto, line 23: Doesn't waiting for an event work?

packages/ui-components/src/components/absolute-time-picker/absolute-time-picker.tsx
```diff
@@ -48,24 +49,39 @@ export class KvAbsoluteTimePicker implements IAbsoluteTimePicker, IAbsoluteTimeP
...
 	handleSelectedRangeDatesChange(value: string[] = []) {
-		// Reset the component state if the selected dates change
 		if (value.length === 0) {
-			this.fromInputValue = '';
-			this.toInputValue = '';
-			this.minDate = '';
-			this.maxDate = '';
+			this.setInputValues('', '');
+			this.setDateLimits('', '');
+		} else {
```
djcouto, line 76: Can we add a return here to avoid the else clause?

packages/ui-components/src/components/relative-time-picker/relative-time-picker.tsx
```diff
@@ -169,6 +178,11 @@ export class KvRelativeTimePicker implements IRelativeTimePicker, IRelativeTimeP
 		return undefined;
 	};
 
+	private getSelectedTimezoneTitle = (): string | undefined => {
```
djcouto, line 181: This function will never return undefined, correct?

djcouto (changes requested):
Last comments 👍

packages/ui-components/src/components/absolute-time-picker/absolute-time-picker.tsx
GoncaloCamaz: Since we now have the reset values methods, the first case would not be needed.
In the second, in my opinion I believe it helps with readability keeping booth values to represent the limits (min and max) but I can change if you think its worth it 👍

djcouto: No need, was just a suggestion.

packages/ui-components/src/components/date-time-input/test/date-time-input.e2e.ts
GoncaloCamaz: It didnt work because of the await before but will update

djcouto (approved):

packages/ui-components/src/components/time-picker/time-picker.helper.ts
```diff
@@ -0,0 +1,138 @@
...
+export const getRelativeTimeLabel = (relativeTimeValue: string | undefined, relativeTimeOptions: IRelativeTimePickerOption[][]): string | undefined =>
+	relativeTimeOptions.flat().find(option => option.value === relativeTimeValue)?.label ?? relativeTimeValue;
+
+export const getTimestampFromDateRange = (range: SelectedTimestampRange, previousTimezone: string, newTimezone: string): SelectedTimestampRange => {
+	const [from, to] = range;
+	// Load date in the previous timezone
+	const parsedFromDate = createFormattedDateFromTimestampInTimezone(from, previousTimezone);
+	const parsedToDate = createFormattedDateFromTimestampInTimezone(to, previousTimezone);
+
+	// Create new timestamps with the updated timezone
```
miguelpinto98, line 114: 👌

miguelpinto98 (approved):
👌

PR #282, opened 2023-04-11
feat(radio-list): add component

pedrorebelocosta:
image


packages/ui-components/src/components/radio-list-item/radio-list-item.helper.tsx
```diff
@@ -0,0 +1,29 @@
+import { JSX, h } from '@stencil/core';
+import { LINK_MATCHER_REGEX } from './radio-list-item.config';
+import { EAnchorTarget } from '../../types';
+
+export const buildDescription = (description?: string): JSX.Element[] => {
```
djcouto, line 5: Are we going forward with this approach? Seems very error prune.
Besides, what if want some part of the description to be bold? How would we achieve that?
Seems to me that a slots solution, despite requiring more work from the developer, would fit every use case we would need.

djcouto (changes requested):
I left one question. Apart from that, LGTM 👍

packages/ui-components/src/components/radio-list/radio-list.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+	render() {
+		return (
+			<Host>
+				<div class="radio-list-container">
+					{this.label && <kv-form-label label={this.label} required={this.required} />}
+					{this.options.map(item => {
+						return (
+							<kv-radio-list-item
+								{...item}
```
miguelpinto98, line 37: what item has more to destruct?

packages/ui-components/src/components/radio-list-item/radio-list-item.helper.tsx
```diff
@@ -0,0 +1,29 @@
...
+	if (!description) return;
+
+	const linkMatches = [...description.matchAll(LINK_MATCHER_REGEX)];
+	const parsedJSX: JSX.Element[] = [];
+	let curIdx = 0;
+
+	for (const match of linkMatches) {
+		const beforeLink = description.slice(curIdx, match.index);
+		if (beforeLink.length > 0) {
```
miguelpinto98, line 14: It is a problem if we push an empty string for `parsedJSX`?

packages/ui-components/src/components/radio-list-item/radio-list-item.helper.tsx
miguelpinto98, line 5: Probably we can export this util to be used in other situations like this?

miguelpinto98 (changes requested):
Almost done! Only check my questions here. 🚀

packages/ui-components/src/components/radio-list-item/radio-list-item.helper.tsx
djcouto: Why don't we use something like this?

https://github.com/markedjs/marked
https://github.com/probablyup/markdown-to-jsx

packages/ui-components/src/components/radio-list-item/radio-list-item.config.ts
```diff
@@ -0,0 +1 @@
+export const LINK_MATCHER_REGEX = /\[(.+?)\]\((https?:\/\/.+?)\)/g;
```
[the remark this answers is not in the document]
pedrorebelocosta: this matches both `http` and `https`, should we remove the check completely?

packages/ui-components/src/components/radio-list/radio-list.tsx
pedrorebelocosta: `label` and `description`, just these 2 props since it's just 2 props, I'll make them explicit here

packages/ui-components/src/components/radio/radio.tsx
```diff
@@ -76,7 +76,7 @@ export class KvRadio {
 					onClick={this.clickThrottler}
 				>
 					<div tabIndex={this.isChecked || this.isDisabled ? -1 : 0} class="circle">
-						<div class="checked-icon"></div>
+						{this.isChecked && <div class="checked-icon"></div>}
```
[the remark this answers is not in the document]
pedrorebelocosta: You're absolutely right 👍

packages/ui-components/src/components/radio-list-item/radio-list-item.helper.tsx
pedrorebelocosta: as parsedJSX is initialized as an empty array, if you pass an empty string the return value will be an empty array which will not render anything

packages/ui-components/src/components/radio-list-item/radio-list-item.helper.tsx
pedrorebelocosta: As discussed earlier, this bit of code is still attached / used by this component only, moving forward, we'll have a separation of this parsing logic from the component.

pedrorebelocosta: Thank you for the input, moving forward I'll isolate the parsing logic from this component, most likely using one of the aforementioned powerful libs!

miguelpinto98 (approved):
🚀

packages/ui-components/src/components/radio-list/radio-list.scss
```diff
@@ -0,0 +1,10 @@
+@import '../../assets/styles/globals';
+
+.radio-list-container {
+	display: flex;
+	flex-direction: column;
+
+	.radio-list-item {
+		margin-bottom: $spacing-2x;
```
djcouto, line 8: You can use `gap: $spacing-2x` on the `.radio-list-container` to avoid the last `.radio-list-item` having an unnecessary margin-bottom.

packages/ui-components/src/components/radio/radio.base.scss
```diff
@@ -35,10 +35,6 @@
 
 		.circle {
 			border-color: var(--radio-input-default-color);
-
-			.checked-icon {
```
djcouto, line 39: What's the purpose of this change?

djcouto (changes requested):

packages/ui-components/src/components/radio/radio.base.scss
miguelpinto98: already checked with @djcouto

packages/ui-components/src/components/radio-list/radio-list.scss
pedrorebelocosta: that would cause a 4px difference since form-label already has 4px margin bottom, but i'll change it to something simpler :)

djcouto (approved):

PR #285, opened 2023-04-19
fix(modal): set correct topbar height

djcouto:


GoncaloCamaz (approved):

miguelpinto98 (approved):
Thanks 💪

packages/ui-components/src/components/modal/modal.base.scss
```diff
@@ -18,7 +18,7 @@
 	--modal-min-height: 235px;
 	--modal-width: fit-content;
 	--modal-min-width: 400px;
-	--modal-topbar-height: 60px;
+	--modal-topbar-height: 51px;
```
pedrorebelocosta, line 21: is it not 50px? 49px + 1px border bottom?

djcouto: Yes 😄

pedrorebelocosta (approved):

PR #287, opened 2023-04-21
fix(time-picker): sync initial view with show calendar toggle state

GoncaloCamaz:
There was a bug when the show calendar toggle was set to true and the component was opened for the first time. The time picker view didnt match with the toggle state.

It was also added an event that emits the state of the toggle when it changes.

### DIFF
--- a/packages/ui-components/src/components/time-picker/time-picker.tsx
+++ b/packages/ui-components/src/components/time-picker/time-picker.tsx
@@ -44,7 +44,7 @@ export class KvTimePicker implements ITimePicker, ITimePickerEvents {
 	/** @inheritdoc */
 	@Prop({ reflect: true }) disabled?: boolean = false;
 	/** @inheritdoc */
-	@Prop({ reflect: true }) showCalendar?: boolean = false;
+	@Prop({ reflect: true, mutable: true }) showCalendar?: boolean = false;
 	/** @inheritdoc */
 	@Prop({ reflect: true }) selectedTimeOption?: ITimePickerTime;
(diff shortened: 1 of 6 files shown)


miguelpinto98 (approved):

packages/ui-components/src/components/time-picker/time-picker.tsx
```diff
@@ -83,12 +83,19 @@ export class KvTimePicker implements ITimePicker, ITimePickerEvents {
...
+	/** @inheritdoc */
+	@Event() showCalendarStateChange: EventEmitter<boolean>;
 
 	@Watch('selectedTimeOption')
 	handleSelectTimeStateChange(timeState: ITimePickerTime) {
 		this.selectedTimeState = timeState;
 	}
 
+	@Watch('showCalendar')
+	handleShowCalendarChange() {
```
djcouto, line 95: This will not be called when the component is mounted, is this the expected behaviour?

djcouto (approved):

packages/ui-components/src/components/time-picker/time-picker.tsx
GoncaloCamaz: I call the syncShowCalendarViewState in the componentWillLoad lifecycle method as well, thus, covering the mount case. 
However, while testing by setting the prop showCalendar in the storybook interface, I noticed that the state of the prop did not affect the view of the component because we were not watching changes to this prop. Therefore I added this Watch to cover that and possible future cases, although, I believe It will not happen but its done just in case 👍

PR #288, opened 2023-04-26
feat(range): add optional unit label and fix some styles

ricardogoncalves89:
image


packages/ui-components/src/components/range/range.base.scss
```diff
@@ -64,9 +65,8 @@
 }
 
 .select-value {
-	@include kv-font-h1-semibold;
-	height: 48px;
-	top: -68px;
+	@include kv-font-h2-semibold;
+	top: -49px;
```
djcouto, line 69: Can we assign a variable to this value in order to give it some meaning?

djcouto (approved):

GoncaloCamaz (approved):

packages/ui-components/src/components/range/range.tsx
```diff
@@ -71,8 +73,12 @@ export class KvRange implements IRange, IRangeEvents {
 						</span>
 					)}
 					<div class="range-min-max">
-						<span>{this.min}</span>
-						<span>{this.max}</span>
+						<span>
+							{this.min} {this.unit}
```
miguelpinto98, line 77: should we create a new property, like `minLabel`? With this we can pass whatever we want. By default these labels should assing the min and max? What you think?

GoncaloCamaz (approved):

miguelpinto98 (approved):
👍

packages/ui-components/src/components/range/range.tsx
```diff
@@ -38,6 +44,14 @@ export class KvRange implements IRange, IRangeEvents {
 		return parseInt(element.value);
 	};
 
+	private getMinLabel = (): string => {
```
pedrorebelocosta, line 47: can we change these to `get` properties? this would avoid the calls to this function on line 78 and 79

packages/ui-components/src/components/relative-time-picker/test/relative-time-picker.e2e.ts
```diff
@@ -31,7 +31,7 @@ describe('relative-time-picker', () => {
 
 		beforeEach(async () => {
 			page = await newE2EPage();
-			await page.setContent('<kv-relative-time-picker inputWrapperContentVisible="true"></kv-relative-time-picker>');
+			await page.setContent('<kv-relative-time-picker timezoneContentVisible="true" />');
```
pedrorebelocosta, line 34: can you explain this change please?

pedrorebelocosta (changes requested):

packages/ui-components/src/components/relative-time-picker/test/relative-time-picker.e2e.ts
GoncaloCamaz: @pedrorebelocosta I did some changes to the name of the props in the relative time picker in one of the last merges but forgot to update the tests. There was no impact big impact in the tests since they have pass before but to keep coherence, I asked Ricardo if he could do this update since the tests failed in this PR. The failure of the test is not related to this change since, as I said, in my PR the tests passed, but this change corrects the name of the props in the tests.

pedrorebelocosta: alright :D I was out of the loop my bad :)

djcouto (approved):

pedrorebelocosta (approved):

PR #289, opened 2023-04-27
feat(button): add loading state

laraujo7:


packages/ui-components/src/components/action-button/action-button.tsx
```diff
@@ -33,11 +35,15 @@ export class KvActionButton implements IButton, IButtonEvents, IAnchor {
 	/** @inheritdoc */
 	@Event() blurButton: EventEmitter<FocusEvent>;
 
+	@State() loadingState: boolean = false;
```
GoncaloCamaz, line 38: Dont forget that since you are using state you will need to sync this variable with the prop `loading` otherwise, no visible change will be seen.

Maybe a componentWillLoad and then a Watch to sync the prop value. However, I believe here a state is not necessary. What do you think?

GoncaloCamaz (changes requested):
Nice loading design! Just need a small fix in the state/prop sync to work 💪 

Is it only suppose to happen when button is from tertiary type? 
Can you also add a story with this behaviour?

packages/ui-components/src/components/action-button/action-button.tsx
pedrorebelocosta: I agree with @GoncaloCamaz, this is best as a prop as you seem to be tying the loading logic to the click of the button and from what I could gather with this approach you'll not be able to unset the loading styling from the parent, what we want is to be able to set a prop that defines if the button is loading according to some operation state.

pedrorebelocosta (changes requested):

GoncaloCamaz (approved):

packages/ui-components/src/components/action-button-split/action-button-split.tsx
```diff
@@ -23,6 +23,8 @@ export class KvActionButtonSplit implements IActionButtonSplitConfig {
 	/** @inheritdoc */
 	@Prop({ reflect: true }) active: boolean = false;
 	/** @inheritdoc */
+	@Prop({ reflect: true }) loadable: boolean = false;
```
pedrorebelocosta, line 26: should this be `loading`?

pedrorebelocosta (changes requested):

ricardogoncalves89 (changes requested):

pedrorebelocosta (approved):

miguelpinto98 (approved):

PR #294, opened 2023-05-11
feat(text-area): add component

pedrorebelocosta:
Required props:
image

With placeholder:
image

With text:
image

SchemaForm widget:
image


GoncaloCamaz (approved):

packages/ui-components/src/assets/svg-symbols.svg
```diff
@@ -1257,5 +1257,9 @@
 		<path d="M9.6 15.6L11 14.175L8.825 12L11 9.825L9.6 8.4L6 12L9.6 15.6ZM14.4 15.6L18 12L14.4 8.4L13 9.825L15.175 12L13 14.175L14.4 15.6ZM5 21C4.45 21 3.97917 20.8042 3.5875 20.4125C3.19583 20.0208 3 19.55 3 19V5C3 4.45 3.19583 3.97917 3.5875 3.5875C3.97917 3.19583 4.45 3 5 3H19C19.55 3 20.0208 3.19583 20.4125 3.5875C20.8042 3.97917 21 4.45 21 5V19C21 19.55 20.8042 20.0208 20.4125 20.4125C20.0208 20.8042 19.55 21 19 21H5Z"/>
 	</svg>
 
+	<symbol id="kv-notes"
+		xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20">
```
miguelpinto98, line 1261: viewBox should be defined as `"0 0 24 24"`

packages/ui-components/src/components/text-area/text-area.scss
```diff
@@ -0,0 +1,72 @@
...
+.text-area-container {
+	display: flex;
+	gap: $spacing-2x;
+
+	.left-icon {
+		margin-top: $spacing-2x;
+	}
+
+	.text-area {
+
```
miguelpinto98, line 21: `-`

packages/ui-components/src/components/text-area/text-area.scss
```diff
@@ -0,0 +1,72 @@
...
+		@include kelvin-night-scrollbar();
+		display: flex;
+		flex-direction: column;
+		flex: 1;
+
+		.text-area-wrapper {
+			padding: $spacing-2x $spacing-3x;
+			border: 1px solid kv-color('neutral-7');
```
miguelpinto98, line 29: should this border be also defined in a variable?

miguelpinto98 (approved):
👍

packages/ui-components/src/components/text-area/text-area.scss
pedrorebelocosta: I think there's a problem with the linter, this is the result after `lint:scss:fix` in this case, I'll try to see if it clears the linter step when removing the space.

packages/ui-components/src/components/text-area/text-area.scss
pedrorebelocosta: do you mean a scss var or css var/prop? 

I decided not to expose it to keep the styling consistent by not allowing changes to this.

packages/ui-components/src/assets/svg-symbols.svg
pedrorebelocosta: that will cause the icon to not be centered when used, this icon's viewbox is `0 0 20 20`

miguelpinto98 (approved):

PR #299, opened 2023-06-02
fix(icons): replace entity-asset and entity-area icons

GoncaloCamaz:
Screenshot 2023-06-02 at 12 20 16


djcouto (changes requested):
The previous `kv-entity-asset` icon was correct, according to our DS.

What we need is to add a new `kv-asset`.

GoncaloCamaz:
> 

That was not the information I had on the ticket. Will need to check.

pedrorebelocosta (approved):

miguelpinto98 (approved):

PR #300, opened 2023-06-05
fix(action-button): remove padding around button due to loading animation

GoncaloCamaz:
Removes the padding that was added to the action button component due to the loading animation PR. This code its not needed and should be removed.


djcouto (approved):

GoncaloCamaz:
> Can you add a screenshot showing the changes? thanks

No visual changes can be seen here. There was an invisible padding added surrounding the button which was causing some spacing in all the buttons.
Screenshot 2023-06-06 at 16 28 49

pedrorebelocosta (approved):

PR #305, opened 2023-06-15
feat(select): add header actions slot

pedrorebelocosta:


packages/ui-components/src/components/multi-select-dropdown/multi-select-dropdown.scss
```diff
@@ -23,4 +23,16 @@
...
+
+	.selection-clear {
+		@include kv-font-span-regular;
+		user-select: none;
+		color: kv-color('neutral-2');
+		cursor: pointer;
+	
+		&.disabled {
+			color: kv-color('neutral-5');
+			cursor: auto;
```
djcouto, line 35: When the clear selection button is disabled, don't we want to show a disallowed cursor?

packages/ui-components/src/components/single-select-dropdown/single-select-dropdown.scss
```diff
@@ -23,4 +23,16 @@
...
+
+	.selection-clear {
+		@include kv-font-span-regular;
+		user-select: none;
+		color: kv-color('neutral-2');
+		cursor: pointer;
+	
+		&.disabled {
+			color: kv-color('neutral-5');
+			cursor: auto;
```
djcouto, line 35: Same as before, shouldn't it be 'not-allowed'?

packages/ui-components/src/components/select/select.tsx
```diff
@@ -17,64 +16,41 @@ export class KvSelect implements ISelect, ISelectEvents {
...
 	/** @inheritdoc */
-	@Prop({ reflect: true }) selectionClearEnabled?: boolean;
-	/** @inheritdoc */
-	@Prop({ reflect: true }) clearSelectionLabel?: string = SELECT_CLEAR_SELECTION_LABEL;
-	/** @inheritdoc */
 	@Prop({ reflect: true }) minHeight?: string;
 	/** @inheritdoc */
 	@Prop({ reflect: true }) maxHeight?: string;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) hasHeaderActions?: boolean = false;
```
djcouto, line 25: Do we really need this? Isn't this just the presence of a slot?

```ts
private hasActionsSlot = () => {
		return !isNil(this.el.querySelector('[slot="select-header-actions"]'));
}
```

packages/ui-components/src/components/select/select.tsx
```diff
@@ -17,64 +16,41 @@ export class KvSelect implements ISelect, ISelectEvents {
...
 		this.searchChange.emit(event.detail);
 	};
 
-	private onClearSelection = () => {
-		this.clearSelection.emit();
-	};
-
-	private getCustomStyle = () =>
-		omitBy(
+	private get customStyle() {
```
djcouto, line 34: Any particular reason to change this to a get? Or is it just a syntax preference?

djcouto (changes requested):

packages/ui-components/src/components/select/select.tsx
pedrorebelocosta: just syntax preference, can revert if you prefer :D

packages/ui-components/src/components/single-select-dropdown/single-select-dropdown.scss
pedrorebelocosta: I'll check with the design team :)

packages/ui-components/src/components/select/select.tsx
```diff
@@ -17,64 +16,41 @@ export class KvSelect implements ISelect, ISelectEvents {
...
-									onClick={this.onClearSelection}
-								>
-									{this.clearSelectionLabel}
-								</div>
-							)}
-						</div>
-					)}
+					<div class={{ 'select-header-container': this.searchable || this.hasHeaderActions }}>
+						{this.searchable && <kv-search value={this.searchValue} placeholder={this.searchPlaceholder} onTextChange={this.onSearchChange} />}
+						{this.hasHeaderActions && <slot name="select-header-actions" />}
```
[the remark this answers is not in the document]
pedrorebelocosta: Thanks for the feedback, I'll be making some changes to the component! :)

packages/ui-components/src/components/multi-select-dropdown/multi-select-dropdown.scss
pedrorebelocosta: Waiting for confirmation on this from design.

pedrorebelocosta: Confirmed :)

djcouto (approved):

GoncaloCamaz (approved):

miguelpinto98 (approved):

PR #308, opened 2023-06-23
fix(select-multi-options): emit only selected options in map

pedrorebelocosta:


GoncaloCamaz (approved):

packages/ui-components/src/components/select-multi-options/select-multi-options.tsx
```diff
@@ -63,8 +63,12 @@ export class KvSelectMultiOptions implements ISelectMultiOptionsConfig, ISelectM
 	};
 
 	private onOptionSelection = ({ detail: option }: CustomEvent<string>) => {
-		// triple bang for handling false, true and undefined
-		const newOptions = { ...this.selectedOptions, [option]: !!!this.selectedOptions?.[option] };
+		let { [option]: isSelected, ...newOptions } = this.selectedOptions;
+
+		if (!isSelected) {
```
djcouto, line 68: The `selectedOptions` is a `Record<string, boolean>` in which its value can be for example:
```typescript
{
    "option-a": true,
    "option-a": false
}
```

In case the user clicks on "option-a", its value won't be changed. I think I get what you are trying to achieve here, but probably you also need to change this prop type to an array instead of an object.

djcouto (changes requested):

djcouto (approved):

packages/ui-components/src/components/select-multi-options/select-multi-options.tsx
pedrorebelocosta: I agree with you but this would imply a breaking change, can we move this discussion [here](https://github.com/kelvininc/ui-components/issues/311)?

PR #310, opened 2023-06-28
feat(schema-form): add tooltip props to submit button

GoncaloCamaz:
Screenshot 2023-06-28 at 17 33 00


pedrorebelocosta (approved):

djcouto (approved):
This adds the functionality that we pretend, which is adding a tooltip to a submit button on the `SchemaForm` component. 

What do you think about we add a configurable tooltip around the `KvActionButton` component enabling this behaviour on all action buttons? I believe in the future we might want this behaviour on other action buttons, that's why IMO this would be a nice feature to add these components.

What do you guys think?

PR #313, opened 2023-07-04
feat(labels-dropdown)!: add component

pedrorebelocosta:
Closed:

image

Selected Item hovered:
image

Open:
image

Partially Selected Level:
image

Adding new label (projected content):
image


packages/ui-components/src/components/labels-dropdown/labels-dropdown.config.ts
```diff
@@ -0,0 +1,10 @@
+import { EIconName, ITextField } from '../../types';
+
+export const CONTAINER_LEFT_MARGIN = 8;
+
+export const TEXTFIELD_CONFIG: Partial<ITextField> = {
```
pedrorebelocosta, line 5: remove

packages/ui-components/src/components/select-option-multi-level/select-option-multi-level.config.ts
```diff
@@ -0,0 +1 @@
+export const LEVEL_OFFSET_PX = 24;
```
pedrorebelocosta, line 1: this is duplicated above, remove

packages/ui-components/src/components/select-option/select-option.tsx
```diff
@@ -41,17 +54,18 @@ export class KvSelectOption implements ISelectOption, ISelectOptionEvents {
 						'disabled': this.disabled,
```
pedrorebelocosta, line 54: I'm having a look at this so it doesn't break the previous bottom slot projection

pedrorebelocosta: I've exposed a css prop `--select-option-text-flex-direction` that allows for the direction of the text-container to be styled to use either a flex row or column layout.

packages/ui-components/src/components/labels-dropdown-input-item/labels-dropdown-input-item.tsx
```diff
@@ -0,0 +1,33 @@
...
+	@Prop({ reflect: true }) value!: string;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) label!: string;
+
+	/** @inheritdoc */
+	@Event() removeClick: EventEmitter<string>;
+
+	render() {
+		return (
+			<div class="labels-input-item" onClick={e => e.stopPropagation()}>
```
djcouto, line 21: Why do we need the `stopPropagation` here?

packages/ui-components/src/components/labels-dropdown-input-item/labels-dropdown-input-item.tsx
```diff
@@ -0,0 +1,33 @@
...
+	shadow: true
+})
+export class KvLabelsDropdownInputItem implements LabelsInputItem, LabelsInputItemEvents {
+	/** @inheritdoc */
+	@Prop({ reflect: true }) value!: string;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) label!: string;
+
+	/** @inheritdoc */
+	@Event() removeClick: EventEmitter<string>;
```
djcouto, line 17: This should be `clickItem`. The remove logic is related to the parent. Furthermore, to make this component reusable we should enable the icon to be configurable and export the icon part.

packages/ui-components/src/components/labels-dropdown/labels-dropdown.scss
```diff
@@ -0,0 +1,68 @@
...
+	.action-label,
+	.selected-items-counter {
+		color: kv-color('neutral-2');
+	}
+
+	.actions {
+		display: flex;
+		flex-direction: row;
+		align-items: center;
+		gap: 8px;
```
djcouto, line 51: `$spacing-2x`

packages/ui-components/src/components/labels-dropdown/labels-dropdown.scss
```diff
@@ -0,0 +1,68 @@
...
+	align-items: center;
+	height: 100%;
+	margin-left: $spacing;
+	justify-content: space-between;
+	gap: $spacing-8x;
+
+	.selected-labels {
+		overflow: hidden;
+		display: flex;
+		gap: 8px;
```
djcouto, line 14: `$spacing-2x`

packages/ui-components/src/components/labels-dropdown/labels-dropdown.scss
```diff
@@ -0,0 +1,68 @@
+@import '../../assets/styles/globals';
+
+.selected-labels-container {
+	display: flex;
+	align-items: center;
+	height: 100%;
+	margin-left: $spacing;
```
djcouto, line 7: Is this correct?

packages/ui-components/src/components/labels-dropdown/labels-dropdown.config.ts
```diff
@@ -0,0 +1,2 @@
+export const CONTAINER_LEFT_MARGIN = 8;
```
djcouto, line 1: The margin is 4px, why are we using 8px here?

packages/ui-components/src/components/labels-dropdown/labels-dropdown.tsx
```diff
@@ -0,0 +1,196 @@
...
+	searchValueWatcher(newValue: string, oldValue: string) {
+		if (newValue === oldValue) return;
+		this._searchValue = newValue;
+	}
+
+	private dropdownElRef: HTMLKvDropdownElement;
+	private flattenedOptionsHash = getFlattenedOptionsMap(this.options);
+
+	private setOverflowingLabels = (): void => {
+		const containerWidth = this.dropdownElRef.querySelector('.selected-labels-container').clientWidth - CONTAINER_LEFT_MARGIN;
```
djcouto, line 70: I believe we should avoid having CSS mixed with JS. But if there's really no other way at least we must use CSS properties to have these values always synced.

packages/ui-components/src/components/labels-dropdown/labels-dropdown.types.ts
```diff
@@ -0,0 +1,19 @@
...
+	/** (Optional) Allows outside implementation to open/close dropdown on demand */
+	isOpen?: boolean;
+}
+
+export interface ILabelsDropdownEvents extends ISelectMultiOptionsEvents {
+	/** Emits a signal whenever the 'select all' action is clicked */
+	selectAll: EventEmitter<void>;
+}
+
+export enum ELabelVisibility {
```
djcouto, line 16: This is css... I don't think we need an enum for this.

packages/ui-components/src/components/text-field/text-field.base.scss
```diff
@@ -40,9 +39,8 @@
 	--text-color-help-text-error: #{kv-color('error')};
 
 	--border-color-error: #{kv-color('error')};
-	--right-slot-padding-right: #{$spacing-9x};
 	--right-slot-width: 36px;
-	--left-slot-padding-left: #{$spacing-6x};
+	--left-slot-width: 32px;
```
djcouto, line 43: this is breaking, right?

djcouto (changes requested):

packages/ui-components/src/components/labels-dropdown-input-item/labels-dropdown-input-item.scss
```diff
@@ -0,0 +1,26 @@
...
+	display: block;
+}
+
+.labels-input-item {
+	height: 16px;
+	padding: $spacing $spacing-3x;
+	background-color: kv-color('primary');
+	display: flex;
+	align-items: center;
+	gap: 8px;
```
GoncaloCamaz, line 13: Use spacing-2x here?

packages/ui-components/src/components/labels-dropdown-input-item/labels-dropdown-input-item.scss
```diff
@@ -0,0 +1,26 @@
...
+
+	.label {
+		@include kv-font-label-small-regular;
+
+		white-space: nowrap;
+		color: kv-color('neutral-0');
+	}
+
+	kv-icon::part(icon) {
+		cursor: pointer;
```
GoncaloCamaz, line 24: Should we here use the include mixin?

packages/ui-components/src/components/labels-dropdown-input-item/labels-dropdown-input-item.types.ts
```diff
@@ -0,0 +1,13 @@
+import { EventEmitter } from '@stencil/core';
+
+export interface LabelsInputItem {
+	/** (Required) The value this input item represents */
+	value: string;
```
GoncaloCamaz, line 5: Shouldn't we here add the required sign? ( value!: string; )

packages/ui-components/src/components/labels-dropdown-input-item/labels-dropdown-input-item.tsx
GoncaloCamaz, line 21: You could use a function to handle the icon click and also add the stop the propagation there. Unless you are doing this to reduce code

packages/ui-components/src/components/labels-dropdown/labels-dropdown.helper.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const getTextFieldConfig = (optionsEmpty: boolean) => {
+	return {
+		placeholder: optionsEmpty ? 'Click to select or create labels...' : '',
+		icon: EIconName.Datalabel
+	};
+};
+
+export const getFlattenedOptionsMap = (options: ISelectMultiOptions): Record<string, ISelectMultiOption> => {
+	if (!options || Object.values(options).length <= 0) {
```
GoncaloCamaz, line 13: == 0 would be enough no?

packages/ui-components/src/components/labels-dropdown/labels-dropdown.scss
GoncaloCamaz, line 14: $spacing-2x here?

packages/ui-components/src/components/labels-dropdown/labels-dropdown.tsx
```diff
@@ -0,0 +1,196 @@
...
+	@Watch('searchValue')
+	searchValueWatcher(newValue: string, oldValue: string) {
+		if (newValue === oldValue) return;
+		this._searchValue = newValue;
+	}
+
+	private dropdownElRef: HTMLKvDropdownElement;
+	private flattenedOptionsHash = getFlattenedOptionsMap(this.options);
+
+	private setOverflowingLabels = (): void => {
```
GoncaloCamaz, line 69: Do you think its worth moving some code here on this function to an util?

packages/ui-components/src/components/labels-dropdown/labels-dropdown.tsx
```diff
@@ -0,0 +1,196 @@
...
+	};
+
+	private renderSelectedOptionsLabels = (): HTMLKvLabelsDropdownInputItemElement[] => {
+		return Object.keys(this.selectedOptions)
+			.filter(value => this.selectedOptions[value])
+			.map((value, idx) => {
+				const selectedOption = this.flattenedOptionsHash[value];
+				if (selectedOption) {
+					const { label } = selectedOption;
+					return <kv-labels-dropdown-input-item key={idx} value={value} label={label} onRemoveClick={this.onOptionRemove} />;
```
GoncaloCamaz, line 109: Should we use the label as key instead of id?

packages/ui-components/src/components/labels-dropdown/labels-dropdown.types.ts
```diff
@@ -0,0 +1,19 @@
...
+type ExcludedSelectMultiOptionsProps = 'selectionClearable' | 'clearSelectionLabel';
+
+export interface ILabelsDropdown extends Omit<ISelectMultiOptionsConfig, ExcludedSelectMultiOptionsProps> {
+	/** (Optional) Allows outside implementation to open/close dropdown on demand */
+	isOpen?: boolean;
+}
+
+export interface ILabelsDropdownEvents extends ISelectMultiOptionsEvents {
+	/** Emits a signal whenever the 'select all' action is clicked */
+	selectAll: EventEmitter<void>;
```
GoncaloCamaz, line 13: Selection cleared is missing isnt it?

GoncaloCamaz (changes requested):
Nice work here!! Just some small questions/suggestions 💪

packages/ui-components/src/components/labels-dropdown/labels-dropdown.types.ts
pedrorebelocosta: inherited from `ISelectMultiOptionsEvents` but fired from a click on a different button

packages/ui-components/src/components/labels-dropdown/labels-dropdown.helper.ts
pedrorebelocosta: indeed :D

packages/ui-components/src/components/labels-dropdown-input-item/labels-dropdown-input-item.tsx
pedrorebelocosta: The `kv-dropdown` component captures click events on the text-field through the `onClick` event, clicks on these labels were bubbling up and causing the dropdown to open when there was a click on one of these items, hence the need to stop event propagation, this seemed like the simplest way of handling this case but let me know if you have any other suggestions ;)

packages/ui-components/src/components/labels-dropdown/labels-dropdown.scss
pedrorebelocosta: It is, I assume you made this comment based on the one below ;) I've assigned the correct value to the const :)

packages/react-ui-components/src/stories/Components/LabelsDropdown.stories.tsx
```diff
@@ -0,0 +1,163 @@
...
+	}, []);
+
+	const onOptionsSelected = useCallback(({ detail }: CustomEvent<Record<string, boolean>>) => {
+		setSelectedOptions(detail);
+	}, []);
+
+	const onAddNewLabel = ({ detail }: CustomEvent<string>) => {
+		const existOption = get(options, ['recents', detail]);
+		if (!existOption) {
+			setOptions(prev => ({
```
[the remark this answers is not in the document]
pedrorebelocosta: using the `merge` function I can't re-order the object keys as intended in a one liner, I can copy from `prev` but this would defeat the purpose of using merge no?

packages/ui-components/src/components/text-field/text-field.base.scss
pedrorebelocosta: since it's a change to the name of this CSS prop, yes, it is breaking. when it comes to how this component is rendered, it should be the same :)

pedrorebelocosta: added breaking notation on commit ;)

packages/react-ui-components/src/stories/Components/LabelsDropdown.stories.tsx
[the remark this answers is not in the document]
djcouto: I didn't see that you were reordering the object keys. Nvmd 👍

GoncaloCamaz (approved):

djcouto (approved):

PR #321, opened 2023-07-19
fix(time-picker)!: cover utc timezone name

GoncaloCamaz:
BREAKING CHANGES: Relative time picker timezones changed from the names to the array of timezone information containing label, value and offset to avoid unnecessary calculations that were being already done in its parent component.


packages/ui-components/src/components/relative-time-picker/relative-time-picker.tsx
```diff
@@ -167,17 +160,13 @@ export class KvRelativeTimePicker implements IRelativeTimePicker, IRelativeTimeP
 		this.timezoneInputClicked.emit(true);
 	};
 
-	private getSelectedTimezone = (): string | undefined => {
+	private getSelectedTimezone = (): string => {
 		if (this.selectedTimezone !== undefined) {
 			return this.selectedTimezone;
 		}
 
 		const defaultTimezone = getDefaultTimezone();
```
djcouto, line 168: `return getDefaultTimezone();`

djcouto (approved):
🚀

ricardogoncalves89 (approved):

pedrorebelocosta (approved):

PR #323, opened 2023-07-20
feat(dropdown)!: add multi level select option to single select dropdown

djcouto:
#### Summary
This PR adds the ability to have multi-level select options in the `single-select-dropdown` component. Besides that, it also allows an option to only be displayed for presentation purposes with the new “presenter” prop of the `kv-select-option` component. This would allow us to remove the “kv-group” component. Furthermore, we no longer require the `kv-multi-level-select-option` since its logic was moved to inside the `kv-select-option` therefore it was also removed.

#### Breaking changes
 - `kv-group` was removed. Please use the `kv-select-option` "options" and "presenter" props to achieve the same behaviour. For example:
```
const options= {
	“option-a”: {
		value: “option-a”,
		label: “Option A”,
		group: “Options”
	},
	“option-b”: {
		value: “option-b”,
		label: “Option B”,
		group: “Options”
	}
}
```
becames
```
const options= {
	“options”: {
		value: “option”,
		presenter: true,
		label: “Options”,
		options: {
			“option-a”: {
				value: “option-a”,
				label: “Option A”,
			},
			“option-b”: {
				value: “option-b”,
				label: “Option B”,
			}
		}
	}
```

 - `kv-select-option-multi-level` component was removed. Please use the `kv-select-option` component instead.


packages/ui-components/src/components/select-multi-options/select-multi-options.helper.ts
```diff
@@ -1,16 +1,56 @@
...
+	const [firtChildrenState, ...otherChildrenStates] = childrenStates;
+	return otherChildrenStates.reduce<EToggleState>((accumulator, childrenState) => {
+		if (accumulator !== childrenState) {
+			return EToggleState.Indeterminate;
 		}
-		return acc;
-	}, {});
+
+		return accumulator;
+	}, firtChildrenState);
```
pedrorebelocosta, line 48: typo

pedrorebelocosta (approved):

GoncaloCamaz (approved):

Insatisfeito (approved):

ricardogoncalves89 (approved):
Awesome job.👌 Please take attention to group selection after a search and selection order in Labels dropdown

packages/ui-components/src/utils/select.helper.ts
```diff
@@ -0,0 +1,56 @@
...
+			accumulator[optionKey] = selectedOption;
+		}
+
+		return accumulator;
+	}, {});
+};
+
+export const getFlattenSelectOptions = (selectOptions: Record<string, ISelectOption> = {}): Record<string, ISelectOption> =>
+	Object.values(selectOptions).reduce<Record<string, ISelectOption>>(
+		(accumulator, selectOption) => ({ ...accumulator, [selectOption.value]: selectOption, ...getFlattenSelectOptions(selectOption.options) }),
```
miguelpinto98, line 36: if we have too many options this process of creating new references in each iteration has a high price.

packages/ui-components/src/utils/select.helper.ts
```diff
@@ -0,0 +1,56 @@
...
+	}, {});
+
+export const buildAllOptionsSelected = (options: ISelectMultiOptions = {}): Record<string, boolean> =>
+	Object.keys(options).reduce<Record<string, boolean>>((accumulator, childKey) => {
+		accumulator[childKey] = true;
+		return accumulator;
+	}, {});
+
+export const buildNoOptionsSelected = (options: ISelectMultiOptions = {}): Record<string, boolean> =>
+	Object.keys(options).reduce<Record<string, boolean>>((accumulator, childKey) => {
```
miguelpinto98, line 53: Do we really need to iterate over all options when we deselect all items? I do not pass an empty object?

packages/react-ui-components/src/stories/Components/MultiSelectDropdown.stories.tsx
```diff
@@ -60,9 +60,10 @@ const MultiSelectDropdownTemplate: ComponentStory<typeof KvMultiSelectDropdown>
 	const filteredOptions = useMemo(() => searchDropdownOptions(searchValue ?? '', options ?? {}), [searchValue, options]);
 
 	const onSelectAll = useCallback(() => {
-		const newSelectedOptions = Object.keys(options ?? {}).reduce<Record<string, boolean>>((acc, cur) => {
-			acc[cur] = true;
-			return acc;
+		const newSelectedOptions = Object.keys(selectHelper.getSelectableOptions(options)).reduce<Record<string, boolean>>((accumulator, optionKey) => {
```
miguelpinto98, line 63: buildAllOptionsSelected ?

packages/ui-components/src/components/select-option/select-option.types.ts
```diff
@@ -11,12 +17,16 @@ export interface ISelectOption {
...
 	/** (optional) If `true` the item is selected */
 	selected?: boolean;
-	/** (optional) If `true` the item is in an indeterminate state (partially selected) */
-	indeterminate?: boolean;
 	/** (optional) If `true` the item is togglable */
 	togglable?: boolean;
-	/** (optional) If `true` styles the item to fit content below label */
-	hasBottomSlot?: boolean;
+	/** (optional) If `true` the item is only for presenting and cannot be selected. */
+	presenter?: boolean;
```
miguelpinto98, line 23: Presenter doesn't seem like a good property name, I not something like selectable?

packages/ui-components/src/components/select-multi-options/select-multi-options.helper.ts
```diff
@@ -1,16 +1,56 @@
...
+		return EToggleState.None;
+	}
+
+	const childrenStates = children.map(childrenOpt => getOptionToggleState(childrenOpt, selectedOptions));
+
+	if (childrenStates.every(childrenState => childrenState === EToggleState.Selected)) {
+		return EToggleState.Selected;
+	}
+
+	if (childrenStates.every(childrenState => childrenState === EToggleState.Indeterminate)) {
```
miguelpinto98, line 33: isn't enough 1 item to be `Indeterminated` to this return true? I said this because in the previous condition, we check all the items are selected, so this one seems more redundant.

djcouto: Yes, I'll do a little refactor here, thank you.

packages/ui-components/src/components/select-option/select-option.types.ts
djcouto: Selectable does seems clearer, will rename it.

packages/react-ui-components/src/stories/Components/MultiSelectDropdown.stories.tsx
djcouto: Nice one, thank you 🙏

packages/ui-components/src/utils/select.helper.ts
djcouto: No, I'll remove this 👍

packages/ui-components/src/components/select/select.tsx
```diff
@@ -68,11 +68,15 @@ export class KvSelect implements ISelect, ISelectEvents {
...
+		const element = this.el.querySelector('[slot="select-header-actions"]');
+
+		return !isNil(element) && element.clientHeight > 0 && element.clientWidth > 0;
 	}
 
 	private get hasLabelSlot() {
-		return !isNil(this.el.querySelector('[slot="select-header-label"]'));
+		const element = this.el.querySelector('[slot="select-header-label"]');
+
+		return !isNil(element) && element.clientHeight > 0 && element.clientWidth > 0;
```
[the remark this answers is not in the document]
djcouto: Absolutely, I will do it 👍

packages/ui-components/src/utils/select.helper.ts
djcouto: Will change this too.

miguelpinto98 (approved):
🚀

PR #333, opened 2023-08-14
feat(schema-form): allow configure the behavior for the default values and a default help text

ricardogoncalves89:
image

Add the `applyDefaults` prop to the `KvSchemaForm` that allow control how empty object fields and array fields where `minItems` is set are populated (default: `All`):

> - `All`: Legacy behavior - set default when there is a primitive value, an non-empty object field, or the field itself is required;
> 
> - `RequiredOnly`: Only sets default when a value is an object and its parent field is required, or it is a primitive value and it is required;
> 
> - `Never`: Does not set defaults;
> 

Add possibility to enable (through the `uiSchema` config) the default helper text:

> Two new configs are available: 
> 
> - `ui:showDefaultValueHelper` [boolean]: show or hide a field helper text with default value (default: false); 
> 
> - `ui:defaultValueHelperPrefix` [text]: prefix label added in the default helper text (default: "Default value is: " )


pedrorebelocosta (approved):

packages/react-ui-components/package.json
```diff
@@ -52,9 +52,9 @@
 		"@kelvininc/ui-components": "0.0.0",
 		"@monaco-editor/react": "^4.4.6",
 		"monaco-editor": "^0.34.1",
-		"@rjsf/core": "5.0.0-beta.12",
-		"@rjsf/utils": "5.0.0-beta.12",
-		"@rjsf/validator-ajv8": "5.0.0-beta.12",
+		"@rjsf/core": "5.12.0",
```
miguelpinto98, line 55: 👌

miguelpinto98 (approved):
👍

packages/react-ui-components/src/components/SchemaForm/Widgets/index.ts
```diff
@@ -1 +1,24 @@
...
+import SelectWidget from './SelectWidget';
+import CheckboxesWidget from './CheckboxesWidget';
+import ReadOnlyValueWidget from './ReadOnlyValueWidget';
+import EmailWidget from './EmailWidget';
+import DateTimeWidget from './DateTimeWidget';
+import DateWidget from './DateWidget';
+import TextareaWidget from './TextareaWidget';
+import { StrictRJSFSchema, RJSFSchema, FormContextType, RegistryWidgetsType } from '@rjsf/utils';
+
+export function generateWidgets<T = any, S extends StrictRJSFSchema = RJSFSchema, F extends FormContextType = any>(): RegistryWidgetsType<T, S, F> {
```
miguelpinto98, line 11: There is some context that we will extend the `FormContextType` ?

PR #335, opened 2023-08-22
fix(react-ui-components): change copy-icons script to generate svg on the right path

miguelpinto98:
When the package manager defined on the project is `pnpm` the node_modules are symlinks. Since this script uses relative paths, this script didn't work as pretended. To overpass this problem the script now has `CUSTOM_INST` environment variable that will copy the icons based on the project directory calls. Check README notes.


pedrorebelocosta (approved):

packages/react-ui-components/.scripts/copy-icons.js
djcouto, line 1: Should we enable the copy destination folder to be configurable? I suggest we rename the `CUSTOM_INST` to `PUBLIC_DIR` env and perhaps once it's not provided fallback the value to `/public`. IMO, it makes more sense, what do you think?

GoncaloCamaz (approved):

packages/react-ui-components/.scripts/copy-icons.js
miguelpinto98: Is not a question of public dir, is more about here react-ui-components assets are located. I start the script with a `CUSTOM_PATH` variable but this turns the postinstall script on the package json with too much length. 😬  

`"postinstall": "CUSTOM_PATH=./node_modules/@kelvininc/react-ui-components/ node ./node_modules/@kelvininc/react-ui-components/.scripts/copy-icons.js"

So I decided to go with the CUSTOM_INST since the PWD will tell us the current path, and the `PWD_REACT_LIB` the path to react lib.

djcouto: I misunderstood the problem you are solving here, thank you 👍

djcouto (approved):

PR #336, opened 2023-08-24
fix(single-select-dropdown): check for invalid selected option value

pedrorebelocosta:


miguelpinto98 (approved):

packages/ui-components/src/components/single-select-dropdown/single-select-dropdown.tsx
```diff
@@ -168,7 +167,7 @@ export class KvSingleSelectDropdown implements ISingleSelectDropdown, ISingleSel
 	}
 
 	private calculateLabelValue() {
-		if (isNil(this.selectedOption)) {
+		if (isNil(this.selectedOption) || isEmpty(this.selectOptions) || values(this.selectOptions).every(isEmpty)) {
```
miguelpinto98, line 170: Why do we need to iterate over the `selectOptions` to define the label?

GoncaloCamaz (approved):

PR #343, opened 2023-09-13
fix(tab-navigation): observe tab visibility for correct calculation of selected indicator width

pedrorebelocosta:


packages/ui-components/src/components/tab-navigation/tab-navigation.tsx
```diff
@@ -43,57 +44,58 @@ export class KvTabNavigation {
...
-		if (selectedTabEl) {
-			this.updateTabIndicatorConfig(getRelativeClientRect(selectedTabEl));
-		}
-	}
+			this.selectedTabIndicatorConfig = {
+				left: `${rect.left}px`,
+				width: `${rect.width}px`
+			};
+		});
+		observer.disconnect();
```
djcouto, line 82: Seems weird disconnecting the observer here, doesn't it?
Is this because the "bug" only happens on the first load?

djcouto (approved):

GoncaloCamaz (approved):

packages/ui-components/src/components/tab-navigation/tab-navigation.tsx
pedrorebelocosta: this is also fired when there's a change in `selectedTabKey` so we can be certain that the element is already visible and a reliable left offset / width calculation can be made.

PR #344, opened 2023-09-14
feat!: add keyboard shortcuts to dropdown components

djcouto:
BREAKING CHANGES:
     - `ISingleSelectDropdownOptions` was renamed to `ISelectSingleOptions`
     - `ISingleSelectDropdownOption` was renamed to `ISelectSingleOption`


packages/react-ui-components/src/stories/Components/SingleSelectDropdown.stories.tsx
```diff
@@ -44,6 +45,12 @@ export default {
 		inputSize: {
 			control: { type: 'radio' },
 			options: Object.values(EComponentSize)
+		},
+		counter: {
+			control: { type: 'boolean ' }
```
GoncaloCamaz, line 50: 'boolean'

GoncaloCamaz (approved):

packages/ui-components/src/components/multi-select-dropdown/multi-select-dropdown.tsx
```diff
@@ -87,63 +94,92 @@ export class KvMultiSelectDropdown implements IMultiSelectDropdown, IMultiSelect
 	/** The Host's element reference */
 	@Element() el: HTMLKvMultiSelectDropdownElement;
 
-	private selectOption = ({ detail: newOptions }: CustomEvent<Record<string, boolean>>) => {
-		this.optionsSelected.emit(newOptions);
-	};
+	@Watch('options')
+	optionsChangeHandler() {
```
pedrorebelocosta, line 98: I think we can change this to:
```
@Watch('options')
@Watch('selectedOptions')
@Watch('displayValue')
labelValueHandler() {
this.calculateLabelValue();
}
```

packages/ui-components/src/components/single-select-dropdown/single-select-dropdown.types.ts
```diff
@@ -39,27 +31,23 @@ export interface ISingleSelectDropdown extends ICustomCss {
...
+	/** (optional) A reference to the dropdown action element */
+	actionElement?: HTMLElement;
+	/** (optional) The object with the dropdown options */
+	options?: ISelectSingleOptions;
+	/** (optional) The object with the dropdown options filtered */
+	filteredOptions?: ISelectSingleOptions;
 }
 
-export interface ISingleSelectDropdownEvents {
+export interface ISingleSelectDropdownEvents extends Omit<ISelectMultiOptionsEvents, 'optionsSelected' | 'selectAll'> {
```
pedrorebelocosta, line 50: should we extend and omit from ISelectEvents instead?

pedrorebelocosta (approved):

packages/ui-components/src/components/single-select-dropdown/single-select-dropdown.types.ts
djcouto: I don't think so otherwise we will need to redefine `optionSelected` that is only present in `ISelectMultiOptionsEvents`.
Nevertheless, after taking a look at this, I noticed there were some unnecessary property definitions. 
I'll remove them 👍

packages/ui-components/src/components/select-multi-options/select-multi-options.tsx
```diff
@@ -193,6 +258,13 @@ export class KvSelectMultiOptions implements ISelectMultiOptionsConfig, ISelectM
 					</slot>
 				)}
 				{this.renderOptions()}
+				{this.shortcuts && (
+					<kv-select-shortcuts-label slot="select-footer">
```
miguelpinto98, line 262: Should this in the future be a slot? I imagine that after this update, they will look at this like a custom thing.

packages/ui-components/src/components/portal/portal.tsx
```diff
@@ -97,8 +97,10 @@ export class KvPortal implements IPortal, IPortalEvents {
 		computePosition(this.reference, this.portal, this.getOptions()).then(({ x, y, placement, middlewareData }) => {
 			if (this.autoUpdate) {
 				const { referenceHidden } = middlewareData.hide;
-				this.visible = !referenceHidden;
-				this.portal.style.zIndex = referenceHidden ? `${PORTAL_Z_INDEX.hidden}` : `${this.zIndex}`;
+				if (this.show) {
```
miguelpinto98, line 100: This here, is it to prevent that dropdown from closing?

miguelpinto98 (approved):
Good work! I only left some questions, but is nothing breaking to request.

packages/ui-components/src/components/portal/portal.tsx
djcouto: Actually is the opposite 😅
This component had an issue when we were controlling the open state from outside, sometimes it opened because the `computePosition` was being called since these portals are never removed from the DOM. This is to prevent that behaviour.

packages/ui-components/src/components/select-multi-options/select-multi-options.tsx
djcouto: You're right, I'll add a slot here 👍

packages/ui-components/src/utils/select.helper.ts
```diff
@@ -52,3 +52,55 @@ export const buildAllOptionsSelected = (options: ISelectMultiOptions = {}): Reco
...
+
+	// Add selected option as a fallback if it exists in the current options
+	const optionsKeys = Object.keys(options);
+	let fallbackOption = optionsKeys[0];
+
+	if (!highlightedOption) {
+		return fallbackOption;
+	}
+
+	const highlightedIndex = optionsKeys.findIndex(name => name === highlightedOption);
```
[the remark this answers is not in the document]
djcouto: I'll give it a try 👌

PR #346, opened 2023-09-21
fix(relative-time-picker): allow set the container min-height

ricardogoncalves89:


packages/ui-components/src/components/relative-time-picker/relative-time-picker.types.ts
```diff
@@ -21,6 +21,12 @@ export interface IRelativeTimePicker {
 	 * if true, the dropdown will be visible; if false, the content will display the timezone title
 	 */
 	timezoneContentVisible?: boolean;
+	/**
+	 * (optional) Additional classes to apply for custom CSS. If multiple classes are
+	 * provided they should be separated by spaces. It is also valid to provide
+	 * CssClassMap with boolean logic.
+	 */
+	customClass?: CustomCssClass;
```
djcouto, line 29: Instead of redeclaring here the prop, change the `IRelativeTimePicker` interface to extend `ICustomCss`.

djcouto (approved):

GoncaloCamaz (approved):

pedrorebelocosta (approved):

miguelpinto98 (approved):

PR #348, opened 2023-09-25
fix(search): show reset icon when input starts with one

GoncaloCamaz:


packages/ui-components/src/components/text-field/text-field.tsx
```diff
@@ -267,13 +272,13 @@ export class KvTextField implements ITextField, ITextFieldEvents {
...
 											'left-slotted': this.hasLeftSlot,
-											'right-slotted': this.hasRightSlot,
+											'right-slotted': !isEmpty(this.actionIcon),
 											'forced-focus': this.focused || this.highlighted
 										}}
 										readonly={this.readonly}
 									/>
 									{this.hasInputContentSlot && (
-										<div class={{ 'input-content-container': true, 'left-slotted': this.hasLeftSlot, 'right-slotted': this.hasRightSlot }}>
+										<div class={{ 'input-content-container': true, 'left-slotted': this.hasLeftSlot, 'right-slotted': !isEmpty(this.actionIcon) }}>
```
miguelpinto98, line 281: I think we can invert the css here since the parent already have the info about the left/rigth-slotted.
So my suggestion:

```
.left-slotted .input-content-container {
	padding-left: var(--left-slot-width);
}

.right-slotted .input-content-container {
	padding-right: var(--right-slot-width);
}
```

And here:
```
<div class="input-content-container">....</div>
```

packages/ui-components/src/components/text-field/text-field.base.scss
```diff
@@ -230,13 +232,13 @@ input {
 		}
 
 		&.disabled {
```
miguelpinto98, line 234: &.disabled kv-icon {
....
}

miguelpinto98 (changes requested):
All good! Only small improvements.

miguelpinto98 (approved):

djcouto (approved):

PR #360, opened 2023-10-25
feat(text-field): add native input focus property

pedrorebelocosta:


pedrorebelocosta:
After a discussion we will be taking a different approach to implement this behavior in the text-field component, for that reason I'm closing this PR.

PR #362, opened 2023-10-30
fix(schema-form): add placeholder on multi select dropdowns

GoncaloCamaz:


pedrorebelocosta (approved):

djcouto (changes requested):
Please extend this behaviour to other fields, like `BaseInputTemplate`

djcouto (approved):

PR #373, opened 2023-11-14
fix(search): disable the auto resize input

ricardogoncalves89:


GoncaloCamaz (approved):

packages/ui-components/src/components/text-field/text-field.tsx
```diff
@@ -75,6 +75,8 @@ export class KvTextField implements ITextField, ITextFieldEvents {
 	@Prop({ reflect: true }) useInputMask?: boolean = false;
 	/** @inheritdoc */
 	@Prop({ reflect: true }) inputMaskRegex?: string = '';
+	/** @inheritdoc */
+	@Prop({ reflect: true }) resizable?: boolean = true;
```
djcouto, line 79: Maybe `fitContent` is a better name. Resizable sounds like the user can resize the input with the cursor 😄

djcouto (approved):
💪

PR #378, opened 2023-11-17
fix(single-select-dropdown): set correct fallback 'selectOption' value

djcouto:


miguelpinto98 (approved):

apply.diff
```diff
@@ -0,0 +1,13 @@
+diff --git a/packages/ui-components/src/components/single-select-dropdown/single-select-dropdown.tsx b/packages/ui-components/src/components/single-select-dropdown/single-select-dropdown.tsx
```
pedrorebelocosta, line 1: 👁️ 👃 👁️

pedrorebelocosta (approved):

pedrorebelocosta (approved):

PR #386, opened 2023-12-11
feat(illustrations): add new illustrations

GoncaloCamaz:
Illustration generation now can handle styling.


packages/ui-components/scripts/generate-illustrations.js
```diff
@@ -59,11 +59,13 @@ async function main() {
 		const className = upperFirst(camelCase(key));
 		const parsedData = parse(defaultItem.data);
 
+		const htmlContent = toHTML(getRoot(parsedData)).replace(/style="mask-type:alpha"/g, 'style={{"mask-type": "alpha"}}');
```
miguelpinto98, line 62: I think we should build a more generic way to handle future styles inside the SVG. So instead of regex this by only this style we should be open to other possible styles and build an object with all the content.

miguelpinto98 (changes requested):

packages/ui-components/scripts/generate-illustrations.js
```diff
@@ -59,15 +60,45 @@ async function main() {
 		const className = upperFirst(camelCase(key));
 		const parsedData = parse(defaultItem.data);
 
+		let htmlContent = toHTML(getRoot(parsedData), { allowDangerousHtml: true });
+	
+		const styleMatches = htmlContent.match(/style="[^"]+"/g) || [];
```
djcouto, line 65: This is due to the "mask-type" style, right? 😞

djcouto (approved):

packages/ui-components/scripts/generate-illustrations.js
GoncaloCamaz: Yes, and now we can handle more styles also if needed.

pedrorebelocosta (approved):

miguelpinto98 (approved):

PR #387, opened 2023-12-14
fix(illustrations): remove added height from illustrations

GoncaloCamaz:
We were having an additional height on the illustrations that was being added by the generated components. This was causing the illustration to add an additional 4px aprox to the specified size we were defining with the custom css props.

### DIFF
--- a/packages/ui-components/src/illustrations/illustrations.scss
+++ b/packages/ui-components/src/illustrations/illustrations.scss
@@ -11,7 +11,8 @@
 
 	width: var(--illustration-width);
 	height: var(--illustration-height);
-
+	display: flex;
+	
 	&-full-size {
 		width: 100%;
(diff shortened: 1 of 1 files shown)


packages/ui-components/src/illustrations/illustration-component.hbs
```diff
@@ -21,6 +21,7 @@ export class {{className}} implements ICustomCss {
 				<svg
 					part="illustration"
 					xmlns="http://www.w3.org/2000/svg"
+					style=\{{ "display": "flex" }}
```
djcouto, line 24: You have an illustration class where you can put this style so you don't need to inline it.

djcouto (approved):

ricardogoncalves89 (approved):

PR #391, opened 2024-01-08
feat: change kv-tag-status icon colors

pedrorebelocosta:
image
image
image
image
image


packages/ui-components/src/assets/svg-symbols.svg
```diff
@@ -1356,4 +1356,14 @@
...
...1 8.70629C4.48975 8.90212 4.72516 9.00004 5.00016 9.00004C5.27516 9.00004 5.51058 8.90212 5.70641 8.70629C5.90225 8.51046 6.00016 8.27504 6.00016 8.00004C6.00016 7.72504 5.90225 7.48962 5.70641 7.29379C5.51058 7.09796 5.27516 7.00004 5.00016 7.00004C4.72516 7.00004 4.48975 7.09796 4.29391 7.29379C4.09808 7.48962 4.00016 7.72504 4.00016 8.00004C4.00016 8.27504 4.09808 8.51046 4.29391 8.70629ZM7.29391 8.70629C7.48975 8.90212 7.72516 9.00004 8.00016 9.00004C8.27516 9.00004 8.51058 8.90212 8.70641 8.70629C8.90225 8.51046 9.00016 8.27504 9.00016 8.00004C9.00016 7.72504 8.90225 7.48962 8.70641 7.29379C8.51058 7.09796 8.27516 7.00004 8.00016 7.00004C7.72516 7.00004 7.48975 7.09796 7.29391 7.29379C7.09808 7.48962 7.00016 7.72504 7.00016 8.00004C7.00016 8.27504 7.09808 8.51046 7.29391 8.70629ZM10.2939 8.70629C10.4897 8.90212 10.7252 9.00004 11.0002 9.00004C11.2752 9.00004 11.5106 8.90212 11.7064 8.70629C11.9022 8.51046 12.0002 8.27504 12.0002 8.00004C12.0002 7.72504 11.9022 7.48962 11.7064 7.29379C11.5106 7.09796 11.2752 7.00004 11.0002 7.00004C10.7252 7.00004 10.4897 7.09796 10.2939 7.29379C10.0981 7.48962 10.0002 7.72504 10.0002 8.00004C10.0002 8.27504 10.0981 8.51046 10.2939 8.70629Z" />
```
miguelpinto98, line 1362: those path are optimized by svgo?

packages/ui-components/src/assets/svg-symbols.svg
```diff
@@ -1356,4 +1356,14 @@
...
 	</symbol>
+	
+	<symbol id="kv-dotted-circle" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16">
```
djcouto, line 1360: This view box must be 24x24. Please update this on both svgs.

djcouto (changes requested):

packages/ui-components/src/components/tag-status/tag-status.scss
djcouto, line 1: This stylesheet seems to be importing the file `../../assets/styles/globals` twice. Can you fix this too?

packages/ui-components/src/assets/svg-symbols.svg
miguelpinto98: not only update this property. It also necessary to resize the SVG to match that viewbox size. If we don't do that we can have a box with 24px but the icon doesn't fit all the box.

packages/ui-components/src/components/tag-status/tag-status.scss
pedrorebelocosta: nice catch 💪  removed double import

packages/ui-components/src/assets/svg-symbols.svg
pedrorebelocosta: I've replaced the icon with the 24x24 version, optimized :)

djcouto (approved):

PR #394, opened 2024-01-15
fix(relative-time-picker): keep timezone search input visible when user searches for values

GoncaloCamaz:
The search input was being hidden when the user searched for values and the number of items was less than the minimum search values.


packages/ui-components/src/components/select/readme.md
```diff
@@ -85,18 +85,6 @@ export const KvSelectExample: React.FC = () => (
 | `"select"` | The select container. |
 
 
-## CSS Custom Properties
```
djcouto, line 88: Revert?

djcouto (approved):

pedrorebelocosta (approved):

PR #400, opened 2024-01-30
feat(copy-to-clipboard): add tooltip config props

djcouto:


GoncaloCamaz (approved):

packages/ui-components/src/components/portal/portal.tsx
```diff
@@ -97,7 +97,7 @@ export class KvPortal implements IPortal, IPortalEvents {
 		computePosition(this.reference, this.portal, this.getOptions()).then(({ x, y, placement, middlewareData }) => {
 			if (this.autoUpdate) {
 				const { referenceHidden } = middlewareData.hide;
-				if (this.show) {
+				if (this.show && this.visible) {
```
miguelpinto98, line 100: This change will not impact other components that use portals?

djcouto: After a few tests, I noticed that it did impact the other portals 😞 
I've made a new commit that with a different approach that seems to be working.

miguelpinto98 (approved):

PR #411, opened 2024-02-23
feat(schema-form): add the radio list widget

ricardogoncalves89:
image


GoncaloCamaz (approved):

packages/react-ui-components/src/components/SchemaForm/Fields/BooleanField/config.ts
```diff
@@ -1,2 +1,7 @@
 export const DEFAULT_TRUE_LABEL = 'True';
```
miguelpinto98, line 1: Those labels are still used?

miguelpinto98 (approved):

PR #413, opened 2024-02-27
fix(time-picker): remove empty group option from timezone list and prevent identical initial and end date

GoncaloCamaz:
This PR addresses two issues in the time picker component.

First we were allowing the users to introduce the same initial and end date if they copied and pasted the values. To fixed this, I added or subtracted one second according with the input if the dates are the same. Also related with this issue, I found that the InputMask package can apply limits correctly only if both limits are defined, so we should always consider applying both limits when using the inputmask with date limits.

The second issue was related with the timezone dropdown. We were letting the users select the parent option (Other Timezones) as a timezone, however, this is regarded as a parent option and should not be selectable. This PR adds the selectable prop to the options build methods to block the user from selecting the parent option.


packages/ui-components/src/components/calendar-advanced-date-selector/calendar-advanced-date-selector.helper.ts
```diff
@@ -48,25 +49,30 @@ export const buildTimezonesDropdownOptions = (timezones: ITimezoneOffset[]): ISe
...
-			}
-		};
+		const timezonesWithoutDefault = filterDefaultTimezone(timezones, timezone?.name);
+		const filteredTimezoneOptions = timezonesWithoutDefault.reduce<ISelectSingleOptions>((accumulator, { label, name }) => {
+			accumulator[name] = accumulator[name] ?? buildDropdownOption(label, name);
+
+			return accumulator;
+		}, {});
+
+		otherTimezoneGroups =
```
GoncaloCamaz, line 59: @djcouto I am also applying this logic here because even though the Other Timezones option is no longer selectable, It was still showing up because we consider the parent option as if it was an option. Since we dont want this option to appear empty, I dont send it if no options are available.

packages/ui-components/src/components/single-select-dropdown/single-select-dropdown.helper.ts
```diff
@@ -1,11 +1,11 @@
-import { isEmpty } from 'lodash';
+import { isBoolean, isEmpty } from 'lodash';
 import { ISelectMultiOption, ISelectSingleOptions } from '../../types';
 
 export const buildSingleSelectOptions = (options: ISelectSingleOptions = {}): Record<string, ISelectMultiOption> =>
 	Object.values(options).reduce<Record<string, ISelectMultiOption>>((accumulator, option) => {
 		accumulator[option.value] = {
 			...option,
-			selectable: isEmpty(option.options),
+			selectable: isBoolean(option.selectable) ? option.selectable : isEmpty(option.options),
```
GoncaloCamaz, line 8: @djcouto I introduced the selectable logic here. If you define the option as selectable or not, we will respect the prop. The fallback is the same as it was before (when we dont define the selectable prop).

miguelpinto98 (approved):

packages/ui-components/src/components/calendar-advanced-date-selector/calendar-advanced-date-selector.helper.ts
GoncaloCamaz: Reverted this changes

djcouto (approved):
🚀

PR #416, opened 2024-03-06
feat(select): add create new option

djcouto:


packages/ui-components/src/components/multi-select-dropdown/multi-select-dropdown.tsx
```diff
@@ -273,6 +273,7 @@ export class KvMultiSelectDropdown implements IMultiSelectDropdown, IMultiSelect
 							onDismiss={this.onDismiss}
 							exportparts="select"
 						>
+							<slot />
```
GoncaloCamaz, line 276: Should we also give a name to this slot?

packages/ui-components/src/components/select-multi-options/select-multi-options.tsx
```diff
@@ -285,6 +285,7 @@ export class KvSelectMultiOptions implements ISelectMultiOptionsConfig, ISelectM
 					</slot>
 				)}
 				{hasCurrentOptions && this.renderOptions()}
+				<slot />
```
GoncaloCamaz, line 288: Here as well?

packages/ui-components/src/components/single-select-dropdown/single-select-dropdown.tsx
```diff
@@ -182,7 +182,9 @@ export class KvSingleSelectDropdown implements ISingleSelectDropdown, ISingleSel
 
 	private selectOption = (selectedOption: string) => {
 		this.optionSelected.emit(selectedOption);
-		this.setOpenState(false);
+		if (!this.selectOptions.flatten[selectedOption].stub) {
```
GoncaloCamaz, line 185: the stub property is optional, should we have precaution here by adding the ? operator?

GoncaloCamaz (approved):

ricardogoncalves89 (approved):

PR #418, opened 2024-03-08
fix(absolute-time-picker)!: handle min and max dates on date click

GoncaloCamaz:
BREAKING CHANGE:
This PR changes the calendarInputMinDate and the calendarInputMaxDate form KvAbsoluteTimePickerDropdown from string to number. It also changes the emitted events type from [] | [number] | [number, number] to only [number] | [number, number]. In the KvTimePicker component, we also changed the emitted event type following the same approach where, the range can only have [number] | [number, number].

A bug was also fixed in the KvAbsoluteTimePicker that happened when the user defined a minimum date and time like 22/12/2024 23:00:00 and then clicked on a date. The clicked date was being set as  22/12/2024 00:00:00, thus, not respecting the minimum date. For the maximum date the same problem was observed, however, the defined date was being set as 22/12/2024 23:59:59.


djcouto (approved):

packages/ui-components/src/components/form-help-text/form-help-text.scss
```diff
@@ -11,7 +11,7 @@
 }
 
 .help-text-container {
-	width: 100%;
+	width: calc(100% - $spacing-2x);
```
GoncaloCamaz, line 14: @ricardogoncalves89 the issue we discussed

PR #421, opened 2024-03-13
fix(date-time-input)!: handle error range exceed in quarter dates

GoncaloCamaz:
Breaking change:

Update KvTimePicker calendarMinDate and calendarMaxDate from string to number to keep consistency with KvAbsoluteTimePickerDropdown. 

This PR fixes an issue that was happening in the date-time-input where the min date and max date were representative of a quarter (1 jan - 31 mar) or other quarters. For some unexpected reason we were having stack range exceed errors in the to input from the calendars. To fix this I removed the watch from min and max and grouped this two props into one. I believe doing so, we are not watching two value changes and thus creating two Inputmask instances but only one instead.


packages/ui-components/src/components/date-time-input/date-time-input.utils.ts
```diff
@@ -4,7 +4,7 @@ import { isEmpty } from 'lodash';
  * This inputmask cannot handle date limits if one of the limit is not defined.
  * Therefore, if we need to use this feature, we should always set the two limits.
  */
-export const getDateLimitsInputmaskConfig = (min?: string, max?: string) => {
+export const getDateLimitsInputmaskConfig = (min?: string, max?: string): Inputmask.Options => {
```
djcouto, line 7: This method could receive the limits array

djcouto (approved):

miguelpinto98 (approved):
👌

PR #430, opened 2024-04-04
fix!(select): change options renderer to be flat to enable virtualization

djcouto:
BREAKING CHANGES:
- The `options` property was removed from `kv-select-option`.
- The `getItemHeight` was removed from `kv-virtualized-list`.


packages/ui-components/src/components/virtualized-list/virtualized-list.tsx
```diff
@@ -17,24 +17,18 @@ export class KvVirtualizedList implements IVirtualizedList {
 	/** @inheritdoc */
 	@Prop({ reflect: true }) getItemKey: (index: number) => string;
 	/** @inheritdoc */
-	@Prop({ reflect: true }) getItemHeight?: (index: number) => number;
-	/** @inheritdoc */
 	@Prop({ reflect: true }) renderItem: RenderItemFunc;
+	/** @inheritdoc */
+	@Prop({ reflect: true }) overscanCount: number = 5;
```
miguelpinto98, line 22: This value is a reference to the children inside of items?

djcouto: This is the number of items to render above/below the visible bounds of the list. This reduces the flickering during scrolling.

miguelpinto98 (approved):
👍

GoncaloCamaz (approved):

ricardogoncalves89 (approved):

PR #433, opened 2024-04-08
fix(toggle-tip): allow toggle tip close when another toggle tip is opened

GoncaloCamaz:
When we have several toggle tips, we should close the ones that are open when we open another one. However, since we were stopping the event propagation, they were not being closed. 
I kept however the same behaviour on the component, allowing this new change through a prop, because we had an old commit adding this stop event propagation. I could not understand the reason for this addition so I kept the old behaviour as default, preventing breaking changes.


packages/ui-components/src/components/toggle-tip/toggle-tip.tsx
```diff
@@ -74,11 +76,12 @@ export class KvToggleTip implements IToggleTip, IToggleTipEvents {
 	};
 
 	private onButtonClick = (ev: MouseEvent) => {
-		ev.stopPropagation();
+		if (this.blockElementClickEventPropagation) ev.stopPropagation();
```
djcouto, line 79: IMO, this should be the default behaviour. 
@ricardogoncalves89 any reason why we were stopping this event propagation? Does this have something to do with this being rendered inside a portal?

djcouto (approved):

miguelpinto98:
The changes introduced on this PR change directly the behavior of this component in previous contexts. Previously on Carbon Maps view, we could have more than 1 tip open at a time.

GoncaloCamaz:
> The changes introduced on this PR change directly the behavior of this component in previous contexts. Previously on Carbon Maps view, we could have more than 1 tip open at a time.

Design team asked to change the previous behaviour. We should close the toggle tip if we click in another.

PR #434, opened 2024-04-11
refactor(calendar)!: remove old calendar components and rename time picker components

GoncaloCamaz:
BREAKING CHANGES

Removed components:
- AdvancedDateSelectDropdown
- CalendarAdvancedDateSelector
- CalendarDay
- Calendar
- CalendarRangeDatesSelector
- CalendarSingleDateSelector
- RangesDateSelectDropdown
- SingleDateSelectDropdown

Renamed Components:
- TimePickerCalendar to Calendar
- TimePickerCalendarDay to CalendarDay


packages/ui-components/src/components/calendar-day/calendar-day.tsx
```diff
@@ -1,53 +1,53 @@
...
-		night: 'calendar-day.night.scss',
-		light: 'calendar-day.light.scss'
-	},
+	styleUrl: 'calendar-day.scss',
 	shadow: true
 })
 export class KvCalendarDay implements ICalendarDay, ICalendarDayEvents {
 	/** @inheritdoc */
-	@Prop({ reflect: true }) day: number;
+	@Prop({ reflect: true }) day!: number;
```
djcouto, line 14: Do you need an exclamation mark? Isn't "required" the default behaviour?

packages/ui-components/src/components/calendar-day/calendar-day.tsx
```diff
@@ -1,53 +1,53 @@
...
+	@Prop({ reflect: true }) disabled?: boolean = false;
 	/** @inheritdoc */
-	@Prop({ reflect: true }) active: boolean = false;
+	@Prop({ reflect: true }) active?: boolean = false;
 	/** @inheritdoc */
-	@Prop({ reflect: false }) inRange: boolean = false;
+	@Prop({ reflect: false }) inRange?: boolean = false;
 	/** @inheritdoc */
-	@Prop({ reflect: false }) leftRounded: boolean = false;
+	@Prop({ reflect: false }) isRangeStartDate?: boolean = false;
```
djcouto, line 22: IMO, the "start date" is not a part of this component and we came up with a different name for this prop.

packages/ui-components/src/components/calendar-day/calendar-day.tsx
```diff
@@ -1,53 +1,53 @@
...
 	/** @inheritdoc */
-	@Prop({ reflect: false }) rightRounded: boolean = false;
+	@Prop({ reflect: false }) isRangeEndDate?: boolean = false;
+	/** @inheritdoc */
+	@Prop({ reflect: false }) isToday?: boolean = false;
+	/** @inheritdoc */
+	@Prop({ reflect: false }) isBetweenSelectedDates?: boolean = false;
```
djcouto, line 28: Isn't this the combination of `isRangeStartDate`, `isRangeEndDate` and `inRange`?

packages/ui-components/src/components/calendar-day/calendar-day.tsx
```diff
@@ -1,53 +1,53 @@
...
+	@Prop({ reflect: true }) active?: boolean = false;
 	/** @inheritdoc */
-	@Prop({ reflect: false }) inRange: boolean = false;
+	@Prop({ reflect: false }) inRange?: boolean = false;
 	/** @inheritdoc */
-	@Prop({ reflect: false }) leftRounded: boolean = false;
+	@Prop({ reflect: false }) isRangeStartDate?: boolean = false;
 	/** @inheritdoc */
-	@Prop({ reflect: false }) rightRounded: boolean = false;
+	@Prop({ reflect: false }) isRangeEndDate?: boolean = false;
```
djcouto, line 24: Same here

packages/ui-components/src/components/calendar-day/readme.md
```diff
@@ -9,16 +11,16 @@
 
 ```html
 <!-- Default -->
-<kv-calendar-day [day]="12"></kv-calendar-day>
+<kv-time-picker-calendar-day [day]="12"></kv-time-picker-calendar-day>
```
djcouto, line 14: I don't think this is correct

packages/ui-components/src/components/calendar-day/calendar-day.types.ts
```diff
@@ -5,36 +5,25 @@ export interface ICalendarDay {
 	day: Number;
 	/** (optional) If `true`, the day is disabled */
 	disabled?: boolean;
-	/** (optional) If `true`, the day is with active style */
+	/** (optional) If `true`, the day is with active */
 	active?: boolean;
-	/** (optional) If `true`, the day is with 'in-range' style */
+	/** (optional) If `true`, the day is in range from a start date to the current hovered day */
```
djcouto, line 10: Is this only related to the "hover" state? Only by looking at this prop name, I assume this also has something to do with the "selected" state.

packages/ui-components/src/components/calendar/calendar.tsx
```diff
@@ -39,18 +36,28 @@ export class KvCalendar implements ICalendar, ICalendarEvents {
...
 	/** @inheritdoc */
+	@Prop({ reflect: false }) hoveredDate?: string;
+	/** @inheritdoc */
 	@Prop({ reflect: false }) disabledDates?: string[] = [];
 	/** @inheritdoc */
 	@Prop({ reflect: false }) minDate?: string;
 	/** @inheritdoc */
 	@Prop({ reflect: false }) maxDate?: string;
+	/** @inheritdoc */
+	@Prop({ reflect: false }) displayPreviousMonthArrow?: boolean = true;
```
djcouto, line 47: `canNavigateToPreviousMonth`

packages/ui-components/src/components/calendar/calendar.tsx
djcouto, line 1: There are a few methods here that don't use class state and therefore belong to an util rather than the component.

djcouto (changes requested):

packages/ui-components/src/components/calendar-day/calendar-day.tsx
GoncaloCamaz: Here the notion of rounded does not apply because the styles for the selected day do not include border radius on the sides. Actually, after taking a better look, I think I dont need two props for this, I only need one like isDayStartOrEndOfRange.

packages/ui-components/src/components/calendar-day/calendar-day.tsx
GoncaloCamaz: No, the inRange represents the dates when we are hovering. In this case we have the initial date selected and we are hovering the days which should represent a certain state. 
The isBetweenSelectedDates represents the state where the start and end date are selected and we set the days in between with another style, which is not the same as the start or end date.

packages/ui-components/src/components/calendar/calendar.tsx
GoncaloCamaz: I don't agree with that name here. This prop is only used to control the presence of the arrow on the left or on the right. The navigation can be enabled or not if the minimum or maximum date has been reached. Therefore, the name can be a little bit deceiving.

packages/ui-components/src/components/calendar/calendar.tsx
GoncaloCamaz: Even though they dont make changes to state, they use a lot of component state variables. Having this methods on the component can avoid having functions with lots of inputs which is not considered a clean code example.

djcouto: I can't entirely agree with you but you're allowed to have your own opinions 🤷‍♂️

djcouto (approved):

packages/ui-components/src/utils/date.helper.ts
```diff
@@ -10,6 +10,7 @@ import isBetween from 'dayjs/plugin/isBetween';
 import quarterOfYear from 'dayjs/plugin/quarterOfYear';
 import utc from 'dayjs/plugin/utc';
 import { TIMEZONES } from './date.config';
+import { SelectedRange } from '../components/calendar/calendar.types';
```
miguelpinto98, line 13: should this SelectedRange be defined in more high level of the apps, seems usage in a lot of places to be categorized on calendar.types

packages/ui-components/src/types.ts
```diff
@@ -22,14 +22,6 @@ export * from './components/tree-item/tree-item.types';
 export * from './components/tree/tree.types';
 export * from './components/badge/badge.types';
 export * from './components/modal/modal.types';
-export * from './components/calendar/calendar.types';
```
miguelpinto98, line 25: hmmm, I have doubts about removing some types of here that can be used inside of the application. Please review the ones that aren't used.

miguelpinto98 (approved):
👍

packages/ui-components/src/types.ts
GoncaloCamaz: Yeah, probably I lost the export with the renaming and forgot to check here! Will take a look 🙏

packages/ui-components/src/utils/date.helper.ts
GoncaloCamaz: Yep, I agree. Some of the types can be in a higher level. I will move this one but I am expecting to move a lot more and have a better understanding about the types location after I start messing with the time picker components.

PR #445, opened 2024-05-10
fix(absolute-time-picker-input): reset input text when user closes dropdown with empty input

GoncaloCamaz:
This PR changes the behaviour of the component to handle the case where a user sets an input and closes the dropdown, and after that, he opens again and cleans the input. If the time was previously set, we should fill the input again with that value. 
The value of the input only changes when the user sets a new valid time or when it picks a new day in the calendar.

### DIFF
--- a/packages/ui-components/src/components/absolute-time-picker-dropdown-input/absolute-time-picker-dropdown-input.tsx
+++ b/packages/ui-components/src/components/absolute-time-picker-dropdown-input/absolute-time-picker-dropdown-input.tsx
@@ -109,20 +109,39 @@ export class KvAbsoluteTimePickerDropdownInput implements IAbsoluteTimePickerDro
 		}
 	};
 
+	private handleEmptyTextChange = () => {
+		if (this.mode === EAbsoluteTimePickerMode.Single) {
+			this.singleTimeInputValue = '';
+		} else {
+			if (this.focusedInput === EInputSource.From) {
+				this.rangeInputValues = { ...this.rangeInputValues, from: '' };
+			} else {
+				this.rangeInputValues = { ...this.rangeInputValues, to: '' };
+			}
+		}
+	};
+
 	private hanleOnTextChange = (event: CustomEvent<string>) => {
 		const parsedDateTime = dayjs(event.detail, DATE_FORMAT);
+
+		if (isEmpty(event.detail)) {
+			this.handleEmptyTextChange();
+			return;
+		}
 
 		if (parsedDateTime.isValid()) {
 			if (this.mode === EAbsoluteTimePickerMode.Single) {
 				this.singleTimeInputValue = parsedDateTime.format(DATE_FORMAT);
+				this.displayedMonth = parsedDateTime;
 				return;
 			}
 
 			if (this.focusedInput === EInputSource.From) {
-				this.rangeInputValues = { from: parsedDateTime.format(DATE_FORMAT), ...this.rangeInputValues };
+				this.rangeInputValues = { ...this.rangeInputValues, from: parsedDateTime.format(DATE_FORMAT) };
 			} else {
-				this.rangeInputValues = { to: parsedDateTime.format(DATE_FORMAT), ...this.rangeInputValues };
+				this.rangeInputValues = { ...this.rangeInputValues, to: parsedDateTime.format(DATE_FORMAT) };
 			}
+			this.displayedMonth = parsedDateTime;
 		}
 	};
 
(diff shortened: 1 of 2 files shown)


packages/ui-components/src/components/absolute-time-picker-dropdown-input/absolute-time-picker-dropdown-input.tsx
```diff
@@ -109,20 +109,39 @@ export class KvAbsoluteTimePickerDropdownInput implements IAbsoluteTimePickerDro
...
 		if (parsedDateTime.isValid()) {
 			if (this.mode === EAbsoluteTimePickerMode.Single) {
 				this.singleTimeInputValue = parsedDateTime.format(DATE_FORMAT);
+				this.displayedMonth = parsedDateTime;
 				return;
 			}
 
 			if (this.focusedInput === EInputSource.From) {
-				this.rangeInputValues = { from: parsedDateTime.format(DATE_FORMAT), ...this.rangeInputValues };
+				this.rangeInputValues = { from: parsedDateTime.format(DATE_FORMAT), to: this.rangeInputValues.to };
```
GoncaloCamaz, line 140: This change here and bellow are necessary for the correct behaviour of the calendar. If I use destructuring here, the component will ignore the new from I am providing. This way I can force the update of the object correctly.

djcouto (approved):

packages/ui-components/src/components/absolute-time-picker-dropdown-input/absolute-time-picker-dropdown-input.tsx
miguelpinto98: Previously, the destructuring wasn't in the wrong order?

```
this.rangeInputValues = { ...this.rangeInputValues, from: parsedDateTime.format(DATE_FORMAT)  };
```

packages/ui-components/src/components/absolute-time-picker-dropdown-input/absolute-time-picker-dropdown-input.tsx
```diff
@@ -138,18 +157,24 @@ export class KvAbsoluteTimePickerDropdownInput implements IAbsoluteTimePickerDro
...
+				this.selectedTimeChange.emit([dayjs(this.singleTimeInputValue, DATE_FORMAT).tz(this.timezone.name, true).valueOf()]);
+			}
 		} else {
-			!isEmpty(this.rangeInputValues?.from) &&
-				!isEmpty(this.rangeInputValues?.to) &&
+			if (isEmpty(this.rangeInputValues?.from) || isEmpty(this.rangeInputValues?.to)) {
+				this.syncSelectedTimeState(this.selectedTime);
+			} else {
 				this.selectedTimeChange.emit([
 					dayjs(this.rangeInputValues.from, DATE_FORMAT).tz(this.timezone.name, true).valueOf(),
```
miguelpinto98, line 174: Hmmm, this implementation can introduce an error in the process if the `this.rangeInputValues.from` or `this.rangeInputValues.to` are empty? Or not because the dayjs on undefined will generate the current day?

GoncaloCamaz: It cannot. I do that validation to check if it is empty. If not I also validate the date with the dayjs isValid method.

GoncaloCamaz: Furthermore, I only emit when the dropdown is closed. If one of the values is empty when the dropdown is closed, the expected behaviour is to set the previously set values because the user has not introduced a new valid date.

packages/ui-components/src/components/absolute-time-picker-dropdown-input/absolute-time-picker-dropdown-input.tsx
```diff
@@ -159,34 +184,59 @@ export class KvAbsoluteTimePickerDropdownInput implements IAbsoluteTimePickerDro
...
+			}
+		} else {
+			const inputDates = getRangeCalendarDates(this.rangeInputValues);
+			if (!isEmpty(inputDates)) {
+				return inputDates;
+			} else if (!isEmpty(this.selectedTime)) {
+				const selectedTimeDates = getRangeInputValues(this.selectedTime, this.timezone.name);
+				return getRangeCalendarDates(selectedTimeDates);
+			}
+		}
```
miguelpinto98, line 222: 😨 

Let's try to beautify this if-else situation:

```
const inputDate = this.mode === EAbsoluteTimePickerMode.Single 
    ? getSingleCalendarDate(this.singleTimeInputValue)
    :  getRangeCalendarDates(this.rangeInputValues);
    
if (inputDate) return inputDate;

if (this.selectedTime) {
   return  this.mode === EAbsoluteTimePickerMode.Single
        ? getSingleCalendarDate(getSingleInputDate(this.selectedTime, this.timezone.name))
	    : getRangeCalendarDates(getRangeInputValues(this.selectedTime, this.timezone.name));
}

return [];
```

miguelpinto98 (approved):

PR #452, opened 2024-06-05
fix(schema-form): remove internal value states

djcouto:


ricardogoncalves89 (approved):

packages/react-ui-components/src/components/SchemaForm/SchemaForm.tsx
```diff
@@ -113,7 +115,6 @@ export function KvSchemaForm<T, S extends StrictRJSFSchema = RJSFSchema>({
 	};
 
 	const discardChanges = () => {
-		setFormData(cloneDeep(submittedData) as T);
```
miguelpinto98, line 116: the new values from discarded option will come from the parent?

miguelpinto98 (approved):

packages/react-ui-components/src/components/SchemaForm/SchemaForm.tsx
djcouto: Yap, that's the point, no internal states.

PR #460, opened 2024-07-05
fix(toaster): watch ttl changes to clear and start new timeout

miguelpinto98:
When the toaster is open and a TTL appears that resets the value to `0` and after a moment appears another data, the toaster doesn't use the default TTL.


danielcarvalhopt (approved):

ricardogoncalves89 (approved):

djcouto (approved):
Are the changes to the readme files intentional?

miguelpinto98:
> Are the changes to the readme files intentional?

Yes, the props are repeated on the files

PR #473, opened 2024-08-29
[WIP]feat(inline-editable-field): remove edit icon and save changes on blur

GoncaloCamaz:
New behaviours were added to this component:
- we should not render the edit icon
- changes should be applied on blur


djcouto (approved):

miguelpinto98 (approved):
👍

PR #474, opened 2024-09-06
fix(inline-editable-field): remove reserved width for edit icon button

GoncaloCamaz:


djcouto (approved):

miguelpinto98:
Some concerns about the flicking right content (if it exists) when edit mode is enabled

ricardogoncalves89 (approved):

GoncaloCamaz:
> Some concerns about the flicking right content (if it exists) when edit mode is enabled

design team wants us to hide what we have on the right of the component when we are editing. Therefore, we will not see things jumping.

PR #477, opened 2024-09-24
fix(select): add ellipsis to items counter

GoncaloCamaz:


packages/ui-components/src/components/select-multi-options/select-multi-options.tsx
```diff
@@ -340,7 +341,9 @@ export class KvSelectMultiOptions implements ISelectMultiOptionsConfig, ISelectM
 				<slot name="select-header-label" slot="select-header-label" />
 				{this.counter && (
 					<div slot="select-header-label">
-						<div class="selected-items-label">Selected: {`${selectedOptionsLength}/${optionsLength}`}</div>
+						<kv-tooltip text={selectedItemsCountText}>
```
djcouto, line 344: We should use the tooltip truncate prop here instead.

djcouto (changes requested):

packages/ui-components/src/components/select-multi-options/select-multi-options.scss
```diff
@@ -21,7 +21,9 @@ kv-select {
 }
 
 .selected-items-label {
+	@include ellipsis();
```
djcouto, line 24: I don't think we still need this after your latest changes.

djcouto (approved):

miguelpinto98 (approved):
Approved, but with one question. 
Does it really make sense to truncate/hide the selected items?

GoncaloCamaz:
> Approved, but with one question. Does it really make sense to truncate/hide the selected items?

The dropdown width is still configured from the outside in case we want to prevent this problem.

PR #493, opened 2024-10-31
feat(dropdown): add action button to select option

djcouto:
#### Screenshot
Screenshot 2024-10-31 at 19 37 50


packages/ui-components/src/components/select-option/select-option.tsx
```diff
@@ -95,11 +98,25 @@ export class KvSelectOption implements ISelectOption, ISelectOptionEvents {
...
+								</div>
+							</div>
+							<div class="right-content">
+								{this.description && <div class="item-description">{this.description}</div>}
+								{this.action && (
+									<div class="action-container">
+										<kv-action-button-icon
+											type={EActionButtonType.Ghost}
+											icon={this.action.icon}
+											onClick={event => event.stopPropagation()}
```
miguelpinto98, line 114: These onClick is necessary to avoid to close the dropdown?

ricardogoncalves89 (approved):

packages/ui-components/src/components/select-option/select-option.tsx
djcouto: Afterall, seem that this is really necessary otherwise the dropdown will think that the option was selected.

miguelpinto98 (approved):

PR #501, opened 2024-11-20
feat: upgrade stencil version

djcouto:


packages/ui-components/src/components/range/readme.md
```diff
@@ -45,12 +45,20 @@ export const KvRangeExample: React.FC = () => (
 | Name                         | Description                        |
 | ---------------------------- | ---------------------------------- |
 | `--range-height`             | Height of the range slider         |
+| `--range-height`             | Height of the range slider         |
```
miguelpinto98, line 48: There are some repeated props on the readme files

packages/ui-components/src/components/inline-editable-field/test/inline-editable-field.e2e.ts
```diff
@@ -33,7 +33,7 @@ describe('KvInlineEditableField (e2e tests)', () => {
 			expect(btns.length).toBe(0);
 		});
 
-		describe('and focus the component', () => {
+		describe.skip('and focus the component', () => {
```
miguelpinto98, line 36: we can't catch the focused element?

packages/ui-components/src/components/inline-editable-field/inline-editable-field.tsx
```diff
@@ -112,9 +112,10 @@ export class KvInlineEditableField {
 		}, DELAYED_BLUR_MS);
 	};
 
-	private handleFocus = () => {
+	public handleFocus = () => {
 		clearTimeout(this.timeoutID);
 
+		console.log('handleFocus');
```
miguelpinto98, line 118: remove?

packages/ui-components/package.json
```diff
@@ -42,13 +42,13 @@
 		"build:watch": "stencil build --watch",
```
miguelpinto98, line 42: `hydrate/` should be added into the files property, right?

packages/react-ui-components/tsconfig.json
```diff
@@ -1,27 +1,21 @@
 {
 	"compilerOptions": {
-		"strict": true,
-		"allowUnreachableCode": false,
-		"allowSyntheticDefaultImports": true,
 		"declaration": true,
+		"noImplicitAny": false,
```
miguelpinto98, line 4: are we sure that we want this config define as false?

miguelpinto98 (changes requested):
Only to have a final check, after the final review

packages/ui-components/src/components/inline-editable-field/test/inline-editable-field.e2e.ts
djcouto: This is a leftover, I will revert this change.

packages/ui-components/src/components/range/readme.md
djcouto: You're right, will revert this changes as well 👍

packages/react-ui-components/tsconfig.json
djcouto: No, this was a leftover as well, will revert this

miguelpinto98 (approved):
🚀

ricardogoncalves89 (approved):

PR #517, opened 2025-02-16
feat(code-editor): export the instance monaco editor reference

ricardogoncalves89:


miguelpinto98:
@ricardogoncalves89 Can you explain a little bit why we need to have this instance?

ricardogoncalves89:
@miguelpinto98 should be able to manage the editor’s markers from outside.

djcouto (approved):

PR #523, opened 2025-03-19
feat: adds more options to Select & Checkbox widget, new icons & minor improvments

miguelpinto98:
…r improvments


ricardogoncalves89 (approved):

packages/ui-components/src/components/tooltip/tooltip.tsx
```diff
@@ -103,7 +103,7 @@ export class KvTooltip implements ITooltip {
 						reference={this.getContentElement()}
 						options={this.getOptions()}
 					>
-						<kv-tooltip-text class={{ ...getClassMap(this.customClass) }} text={this.text} />
+						{!isEmpty(this.text) && <kv-tooltip-text class={{ ...getClassMap(this.customClass) }} text={this.text} />}
```
djcouto, line 106: Do we want to render the div above if there's no content?
Should this condition be checked earlier?

djcouto (approved):
🚀🚀🚀

PR #525, opened 2025-03-19
fix: adds bookmark icon

miguelpinto98:


packages/ui-components/src/assets/svg-symbols.svg
```diff
@@ -1468,4 +1468,8 @@
 	<symbol id="kv-content-expand" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20">
 		<path d="M4.16699 15.8334V10.8334H5.83366V14.1667H9.16699V15.8334H4.16699ZM14.167 9.16669V5.83335H10.8337V4.16669H15.8337V9.16669H14.167Z" />
 	</symbol>
+
+	<symbol id="kv-bookmark" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16">
```
djcouto, line 1472: Can we change this icon to have a view box of 24x24?

djcouto (changes requested):

packages/ui-components/src/assets/svg-symbols.svg
miguelpinto98: Some weeks ago, me and @danielcarvalhopt concluded that isn't necessary to do it. If the icon was designed with 16x16 of viewBox it will adapt to each necessary size.

djcouto: Hum, fine by me 👍

djcouto (approved):

PR #528, opened 2025-04-14
feat(schema-form): only show errors after dirty field

ricardogoncalves89:


packages/react-ui-components/src/components/SchemaForm/Templates/BaseInputTemplate/BaseInputTemplate.tsx
```diff
@@ -25,7 +26,14 @@ const BaseInputTemplate = <T, S extends StrictRJSFSchema = RJSFSchema, F extends
 	formContext,
 	type
 }: BaseInputTemplateProps<T, S, F>) => {
-	const _onChange = useCallback((value: CustomEvent<string>) => onChange(value?.detail ? value.detail : options.emptyValue), [onChange, options]);
+	const { isDirty, setDirty } = useCurrentDirtyFieldsContext();
+	const _onChange = useCallback(
+		(value: CustomEvent<string>) => {
+			setDirty(id);
```
djcouto, line 32: Shouldn't we verify whether the value has changed? Otherwise, we are merely checking if the field was modified, correct? I'm uncertain if that was the intended purpose.

PR #531, opened 2025-04-22
fix(editable-inline): watchs disable state to destroy previous state

miguelpinto98:


packages/ui-components/src/components/inline-editable-field/inline-editable-field.tsx
```diff
@@ -81,6 +81,15 @@ export class KvInlineEditableField {
 		}
 	}
 
+	@Watch('disabled')
+	handleDisableChange(state) {
```
djcouto, line 85: This is not called when this component is "mounted", not sure if we need to add it to `connectedCallback`

djcouto (changes requested):

ricardogoncalves89 (approved):

packages/ui-components/src/components/inline-editable-field/inline-editable-field.tsx
djcouto: Okok 👍

djcouto (approved):

PR #542, opened 2025-05-09
refactor!: remove components light mode version

djcouto:


miguelpinto98:
Are we sure about this? I understand that we don't really use any light component, but having the ui-components prepared to render components with that specificity seems something good.

ricardogoncalves89 (approved):

PR #555, opened 2025-07-08
feat: adds some improvments on absolute calendar and select option

miguelpinto98:
- feat(select-option): adds a tooltip into icon and export icon part to be stylized outside
- feat(absolute-time-picker-dropdown): enables dropdown open state to be controlled externally


packages/ui-components/src/components/select-option/select-option.tsx
```diff
@@ -28,6 +28,8 @@ export class KvSelectOption implements ISelectOption, ISelectOptionEvents {
 	/** @inheritdoc */
 	@Prop({ reflect: true }) icon?: EIconName;
 	/** @inheritdoc */
+	@Prop({ reflect: true }) iconTooltipText?: EIconName;
```
djcouto, line 31: I think this type is not correct. It should be string, right?

djcouto (changes requested):

djcouto (approved):

PR #560, opened 2025-07-30
feat(dropdown): add `select-footer` slot to multi and single select dropdowns

djcouto:


packages/ui-components/src/components/multi-select-dropdown/multi-select-dropdown.tsx
```diff
@@ -305,6 +305,7 @@ export class KvMultiSelectDropdown implements IMultiSelectDropdown, IMultiSelect
 							<slot name="select-header-label" slot="select-header-label" />
 							<slot name="no-data-available" slot="no-data-available" />
 							<slot name="no-results-found" slot="no-results-found" />
+							<slot name="select-footer" slot="select-footer" />
```
miguelpinto98, line 308: should do we create a more agnostic slot to use when we want? 😄

djcouto: Yes, it's a good idea 😅
I'll do that on another PR.

PR #569, opened 2025-09-10
feat(select-option): add right icon tooltip

ricardogoncalves89:


djcouto (changes requested):
Please compile locally to generate the respective updates to the component's `readme.md` file.

PR #615, opened 2026-02-20
feat!: migrate badge, tab-item and tab-navigation to design tokens

ricardogoncalves89:
BREAKING CHANGE: `EBadgeState` was removed and replaced with `EBadgeType` (Primary, Secondary). The `state` prop on `kv-badge` is now `type`, and `disabled` is a new explicit prop instead of a state variant.

BREAKING CHANGE: `kv-action-button-icon` removed `badgeLabel` and `badgeState` props. The badge is no longer part of the action button icon.

BREAKING CHANGE: `kv-tab-item` removed props `size`, `hasNotification`, `notificationColor`, `icon`, and `state`. The tab type is now controlled via `type` (ETabItemType.Primary | Secondary) and slot-based content (`right-slot`) replaces inline icon and notification dot rendering.

BREAKING CHANGE: `kv-tab-navigation` removed `notifications` and `size` props. Use `type` (ETabItemType.Primary | Secondary) to control the visual variant. Badge and tag-status content is now passed via `ITabNavigationItem` fields `badge`, `badgeType`, `tagIcon`, and `tagState`.


apps/react-storybook/src/components/buttons/icon/Icon.stories.tsx
```diff
@@ -2,7 +2,7 @@ import type { Meta, StoryObj } from "@storybook/react";
 
 import {
 	EActionButtonType,
-	EBadgeState,
+	EBadgeType,
```
djcouto: I think this is no longer being used.

djcouto (approved):

PR #618, opened 2026-02-24
feat(switch-button): migrate to design token system

ricardogoncalves89:


djcouto (approved):

packages/ui-components/src/components/switch-button/switch-button.scss
```diff
@@ -112,14 +88,10 @@ $icon-sizes: (
 	display: inline-flex;
 	align-items: center;
 	justify-content: center;
-
+	box-sizing: border-box;
```
miguelpinto98, line 91: this shouldn't exist on the default class? Why we need this here?

miguelpinto98 (approved):

packages/ui-components/src/components/switch-button/switch-button.scss
ricardogoncalves89: We need this now to have the correct aspect with the new sizes that come from the design tokens.

PR #623, opened 2026-02-26
feat(radio, checkbox): migrate to design token system

ricardogoncalves89:
- Refactor kv-radio: replace internal @State with direct @Prop usage, swap circle/checked-icon markup for kv-icon with action-icon slot, add IRadio/IRadioEvents interfaces and export radio.types, update checkedChange event type from boolean to Event
- Refactor kv-checkbox: remove dedicated scss, delegate rendering to kv-radio via action-icon slot, extend IRadio interface, update clickCheckbox event type to Event
- Migrate radio.scss to design tokens: replace hardcoded values with input/radio component_semantics tokens, add hover/pressed/disabled background states via overlay tokens, use clickable/non-clickable mixins
- Update checkbox and radio icon SVGs (check-state, uncheck-state, indeterminate-state) to 24x24 viewBox
- Update select-option to add --with-checkbox modifier, fix hover selector specificity, and pass EComponentSize.Small to kv-checkbox
- Update radio-list-item to use spacing tokens, add focus state and tabIndex, pass EComponentSize.Large to kv-radio
- Update Storybook stories: add interactive template with useArgs for radio, add size argType to checkbox
- Mark radio and checkbox as migrated in MIGRATION_STATUS.md


packages/ui-components/src/components/checkbox/checkbox.tsx
```diff
@@ -22,7 +23,7 @@ export class KvCheckbox implements ICheckbox, ICheckboxEvents {
 	@Prop({ reflect: true }) indeterminate?: boolean = false;
 
 	/** @inheritdoc */
-	@Event() clickCheckbox: EventEmitter<MouseEvent>;
```
djcouto, line 25: Why do we need this change? It comes in handy to have the original MouseEvent being propagated.

djcouto (approved):

miguelpinto98 (approved):

packages/ui-components/src/components/checkbox/checkbox.tsx
ricardogoncalves89: Because KvRadio can be selected either by a mouse click or by pressing the space key.

PR #624, opened 2026-02-26
feat(link): migrate to design token system

ricardogoncalves89:
- Add leftIcon, rightIcon, disabled and inline props
- Expose label CSS part for external styling
- Use design tokens for colors, spacing and typography across all states (default, hover, focus, visited, disabled) for both standard and inline variants
- Fix icon color in :visited state by using --icon-color: currentColor, working around browser security restrictions that block CSS custom property updates inside :visited
- Update Storybook stories with icon controls and new icon-based story variants

### DIFF
--- a/packages/ui-components/src/components/link/link.scss
+++ b/packages/ui-components/src/components/link/link.scss
@@ -2,25 +2,107 @@
 
 :host {
 	.link-container {
+		width: fit-content;
 		display: flex;
 		flex-direction: column;
 
-		.label {
-			@include body-m-semibold();
-			text-decoration: none;
-			user-select: none;
-			color: kv-color('text');
-			cursor: pointer;
+		&--disabled {
+			@include non-clickable;
+		}
 
-			&:hover {
-				color: kv-color('primary');
+		.label {
+			display: inline-flex;
+			align-items: center;
+			gap: var(--link-gap-default);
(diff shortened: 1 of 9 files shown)


djcouto (approved):

packages/ui-components/src/components/link/link.scss
```diff
@@ -2,25 +2,106 @@
...
+			}
+
+			&:visited:not(.label--disabled) {
+				color: var(--link-label-visited);
+			}
+
+			&:focus:not(.label--disabled),
+			&:hover:not(.label--disabled) {
+				color: var(--link-label-hover);
 				text-decoration: underline;
```
miguelpinto98, line 41: The links that I been using on the app I change this property to border-bottom.
That approach allow us to add a simple animation transition that give more finesse 😄 

Give it a try

miguelpinto98 (approved):

packages/ui-components/src/components/link/link.scss
ricardogoncalves89: Don’t you have problems with the height when the link is inline with other text?

ricardogoncalves89: I added a transition to the `text-decoration`, like we have in the app.

PR #625, opened 2026-03-02
feat(text-field,search,dropdown): migrate to design token system

ricardogoncalves89:


djcouto (approved):

packages/ui-components/src/components/form-help-text/form-help-text.scss
```diff
@@ -6,36 +6,42 @@
...
+	--help-text-left-spacing: var(--spacing-md, #{$spacing-2x});
+
+	display: flex;
 }
 
 .help-text-container {
-	width: calc(100% - $spacing-2x);
+	display: inline-flex;
+	flex-flow: wrap;
+	width: calc(100% - var(--help-text-left-spacing));
```
miguelpinto98: The gap from flex can't be useful here to avoid the width calculation?

miguelpinto98 (approved):

PR #626, opened 2026-03-03
fix(select): preserve options order with numeric-like keys

miguelpinto98:
## Summary
- Fixed JavaScript's behavior of reordering numeric-like string keys (e.g., '98000') in select options
- Changed `currentFlatten` from object to array to preserve insertion order during rendering
- Added `buildSelectOptionsArray()`, `flattenOptionsReducer()`, `getFlattenSelectOptionsArray()`, and `flattenSelectOptionsArray()` helpers
- Added unit tests for the new array-based functions

## Test plan
- [x] Unit tests pass
- [x] Build succeeds
- [ ] Manual test with numeric-like option keys to verify correct ordering

🤖 Generated with [Claude Code](https://claude.com/claude-code)


CLAUDE.md
djcouto, line 1: This should be AGENTS.md, and CLAUDE.md must be created as a symlink to that file.

djcouto (approved):

PR #640, opened 2026-04-13
feat(time-picker): migrate time pickers to  design tokens

ricardogoncalves89:


apps/react-storybook/src/foundation/icons/Icons.scss
```diff
@@ -50,17 +50,17 @@
 		justify-content: center;
 		border: 1px solid #f2f2f2;
 		border-radius: 4px;
```
djcouto: Should we migrate this, too?

djcouto (approved):

PR #674, opened 2026-07-30
fix(text-field): stop numeric input mask from silently rewriting the value

ricardogoncalves89:
Bump inputmask to 5.0.10-beta.71 and tame when the numeric alias enforces min/max so the displayed value never diverges from the value prop:

- programmatic writes (value prop, initial value, re-renders) render as-is instead of being silently clamped to the limits
- out-of-range keystrokes and pastes are still rejected while editing
- on blur, the value left by the user is clamped to the limits and the rewrite is propagated through textChange
- reset the mask state when the field is cleared by deleting a selection (5.0.10 leaves a phantom radix position behind, landing the next digits on the decimal part)
- only sync masked inputs: native input types (number, datetime-local) normalize or reject values internally and adopting their empty value would clear the consumer state and loop with schema-form defaults

SchemaForm number fields now render without the input mask by default; setting useInputMask in the uiSchema still enables it.


packages/react-ui-components/src/components/SchemaForm/Templates/BaseInputTemplate/BaseInputTemplate.tsx
```diff
@@ -87,6 +87,11 @@ const BaseInputTemplate = <T, S extends StrictRJSFSchema = RJSFSchema, F extends
 		[schema.examples, schema.default]
 	);
 
+	const shouldUseInputMask = useMemo(
+		() => (uiSchema.useInputMask === undefined && inputType === EInputFieldType.Number ? false : uiSchema.useInputMask),
+		[uiSchema.useInputMask, inputType]
+	);
```
djcouto, line 93: Can't we simplify this by adding a default assign value to the destructured useInputMask in line 81?

```
	const { componentSize: optionComponentSize, useInputMask = false, inputMaskRegex, valuePrefix, badge } = uiSchema;
```

miguelpinto98 (approved):

PR #679, opened 2026-08-28
feat(time-picker): single-date custom interval, hideable calendar toggle and snooze icon

djcouto:
## Summary

Gives `kv-time-picker` the generic, opt-in capabilities needed for a duration / "pick a future date" picker (Figma: *Assets — Snooze — 1.0*), without adding any feature-specific presets to the library. Consumers compose their own option list with the existing `buildRelativeTimeOption` helper — see the new `FutureDurations` story.

## Changes

**`relative-time` utils**
- New `ERelativeTimeRangeKey.Next_3_D / Next_14_D / Next_30_D / Next_60_D` with matching `RELATIVE_TIME_OFFSETS`.
- `ERelativeTimeRangeKey` added to the React enum bridge (`react-ui-components/src/ui-components.ts`) so React consumers can reference the keys at runtime.

**`kv-time-picker`** (defaults preserve current behaviour)
- `displayCalendarToggle` (default `true`) — hides the "Show Calendar" switch.
- `calendarMode` (default `Range`) — forwards `EAbsoluteTimePickerMode.Single` to the custom-interval calendar; Apply validates against one date and emits a one-element `range`.
- Relative view is now content-sized (capped at the previous 347px, so the default list renders identically) instead of always reserving the maximum height.
- Default `last-24-h` preselection only applies when the provided options contain that key (previously Apply could emit a stale past range for a custom option list).
- `getTimestampFromDateRange` / `handleAbsoluteDatesChange` no longer turn a one-element range into a bogus `[from, now]` pair.
- Cancel button is `Secondary` (per design system).

**Icons** — `EIconName.Snooze` (`kv-snooze`).

**Storybook** — `Time Picker / Time Picker / FutureDurations` (options in `src/mocks/time-picker.ts`), plus `argTypes` for the new props.

**Tests** — new `time-picker/test/time-picker.spec.tsx`; new-key cases in `relative-time.spec.ts`.

## Verification

- `pnpm build:packages`, `pnpm lint`, Prettier, Storybook `tsc`: green.
- Core spec suite: 61 suites / 361 tests pass (incl. the SVGO sprite guard).
- Storybook, driven with Playwright: `FutureDurations` panel renders 328px wide with a 233px list, rows resolve to the target dates, no toggle/timezone row, Cancel bordered, Apply disabled on open → enabled on select; Custom Interval opens the single "Day & Hour" calendar, one click enables Apply, payload is `{ key: 'customize-interval', range: [ts] }`. `DefaultState` unchanged (347px, `Last 24 hours` preselected, scroll intact); FullView divider spans full height. No console errors.

## Notes for reviewers

- The only intentional visible change for existing consumers is the Secondary Cancel button; short option lists now fit their content instead of padding to 347px.
- Panel height is ~10px taller than the Figma frame because the shared `PADDING_SIZE` (16px) differs from the design's 12px slider padding — left as-is deliberately.
- Pre-existing naming quirk left alone (renaming is breaking): `Next_5_M` is minutes while `Next_1_M` is months. The new `_D` keys are unambiguous.


djcouto:
Addressed the automated review comments in `0bab97bb` (amended into the time-picker commit):

- **`validateNewRange` length check** (Copilot): the helper now returns `false` unless `range.length === expectedSize`, so a two-element custom range can no longer pass the single-date check. Test added.
- **Single calendar not initialised from `selectedDates`** (Codex): `kv-absolute-time-picker` gained a `componentWillLoad` that syncs the inputs and displayed month with preselected dates (`@Watch` doesn't run for initial prop values). `initialDate` still takes precedence for the displayed month. Covers range mode too, which had the same gap. Tests added for single, range, and `initialDate` precedence.
- **Readme typos** (Copilot): fixed at the source in `time-picker.types.ts` jsdoc (`otion` → `option`, `doens't` → `doesn't`) and the readme regenerated.

djcouto:
Second round of Copilot comments addressed in `8ac01b52` (amended into the time-picker commit):

- **Truthiness on numeric timestamps** in `getAbsoluteTimePickerRangeDates` and `validateNewRange`: both now use `isNumber` so a `0` (epoch) timestamp is treated as present. Tests added for the epoch case.
- **`getRelativeViewHeight` typing**: the `options` parameter is now `IRelativeTimePickerOption[][] | undefined`, matching the nullish handling (the prop it receives is optional). Test added for `undefined`.

miguelpinto98 (approved):

PR #680, opened 2026-09-01
fix(dropdown): search options locally when no filtered options are given

djcouto:
## Summary

`kv-single-select-dropdown` and `kv-multi-select-dropdown` now filter their own options as the user
types. Search previously did nothing at all unless the consumer wired up filtering by hand, so every
call site had to reimplement the same filter to make a searchable dropdown actually search.

### Problem

Typing in a searchable dropdown left the option list untouched. The dropdowns tracked the search term
internally (`_searchValue`) and forwarded it down to `kv-select-multi-options`, but that component only
ever rendered `filteredOptions ?? options` — so with no `filteredOptions` prop supplied, it rendered the
full, unfiltered list no matter what was typed.

### Root cause

Filtering was never the component's job. `filteredOptions` was the only path to a filtered list, which
made local search opt-in via boilerplate: each consumer had to listen to `onSearchChange`, hold the term
in its own state, run `searchDropdownOptions` itself, and feed the result back in. Miss any part of that
chain and search silently no-ops.

### Fix

`kv-select-multi-options` now runs its own `searchValue` through `searchDropdownOptions` and re-renders on
`searchValue` changes. `filteredOptions` keeps working as an explicit override for external or server-side
filtering, so existing consumers that do their own filtering are unaffected.

## Changes

-   **`fix(select-multi-options)`** — `currentOptions` returns
    `filteredOptions ?? searchDropdownOptions(searchValue, options)`, and `@Watch('searchValue')` was added
    to the rebuild watcher so a new search term recomputes the rendered options. An explicitly passed empty
    `filteredOptions` object still means "no matches" rather than falling back to local search.
-   **`fix(select.helper)`** — `searchDropdownOptions` now defaults `term` to `''`, so it can be called with
    the dropdown's initially-`undefined` `searchValue` and return the untouched options.
-   **`refactor(relative-time-picker)`** — dropped the `timezonesSearchTerm` state, its `@Watch`, the
    `timezoneFilteredDropdownOptions` state, and the `onSearchChange` handler. The timezone dropdown filters
    itself now.
-   **`refactor(schema-form)`** — `SelectWidget` no longer holds a `searchTerm` state or computes
    `filteredOptions`; the local `searchDropdownOptions` helper in `SelectWidget/utils.ts` is gone.
-   **`docs`** — `filteredOptions` is documented as an override ("Externally filtered dropdown options.
    When defined, these override the default local search results.") across the three component readmes and
    the two type definitions.
-   **`test(select-multi-options)`** — new spec covering local filtering, clearing the term, the no-results
    state, and `filteredOptions` taking precedence. `select.helper.spec.ts` gains direct
    `searchDropdownOptions` coverage: empty term, case-insensitive label and value matching, parents keeping
    only matching descendants, and no-match.
-   **`refactor(storybook)`** — the single- and multi-select dropdown story templates drop their manual
    filtering. The `ExternalSearch` story deliberately keeps its wiring, since it demos the override.

## Test Plan

-   [x] `pnpm test:spec` in `packages/ui-components` — 61 suites, 350 tests, 123 snapshots pass, no
        snapshot churn from the new default filtering.
-   [x] `pnpm lint` clean in `packages/ui-components`, `packages/react-ui-components`, and
        `apps/react-storybook`.
-   [ ] E2E suite not run locally — left to CI.
-   Manual: `pnpm build:packages && pnpm storybook`, then
    1. **Single Select Dropdown → Searchable** — type into the search field and confirm the list narrows as
       you type, matching on both label and value, and that clearing the field restores every option.
    2. **Single Select Dropdown → Searchable** — type a term that matches nothing and confirm the
       "no results" state appears.
    3. **Multi Select Dropdown → Searchable** — same, and confirm selections made before searching survive
       filtering and clearing.
    4. **Single Select Dropdown → External Search** — confirm the externally-filtered story still filters
       through its own `filteredOptions`, proving the override path is intact.
    5. **Relative Time Picker** — open the timezone dropdown and confirm its search still filters, now with
       no wiring of its own.
    6. **SchemaForm** — render a form with a searchable `enum` select and confirm search filters the options.


packages/ui-components/src/utils/select.helper.ts
```diff
@@ -144,7 +144,14 @@ export const searchDropdownOptions = (term: string, options: ISelectSingleOption
 			return accumulator;
 		}
 
-		if (isSubString(term, option.label) || isSubString(term, option.value)) {
+		// A blank label is not a searchable identity, so those options are matched by their value alone.
+		if (isValidLabel(option.label) && isSubString(term, option.label)) {
```
miguelpinto98, line 148: 👍

miguelpinto98 (approved):

